Issue #88, opened 2026-04-24
[INFRA] Propagate fault-injection exit code uniformly across all per-module targets


renatosfagundes:
## Context

Review of PR #85 surfaced that the `fault-uds` Makefile target was unconditionally printing `(non-zero exit expected while bugs are pending patch)` even when the fault-injection suite exited 0, and the `tee` pipe was masking the real exit code so CI step state did not reflect whether the suite actually passed or failed.

PR #85 fixes `fault-uds`:
- Saves the binary's exit code into a shell variable;
- Chooses a message conditional on that exit (PASS → "flip `continue-on-error` to false in `vv-uds.yml`"; FAIL → "non-zero exit expected while bugs are pending patch");
- Re-exits with the real code so the workflow step colour matches reality (still non-blocking while `continue-on-error: true` holds at the YAML level).

## The same anti-pattern exists elsewhere

`fault-decision` on `development` has the identical construct (same `tee` + unconditional echo) — when Eryca flipped fault-injection to blocking after her bugs are patched, the echo will claim "non-zero exit expected" even on PASS runs.

Every subsequent per-module V&V PR on the way (`fault-can` in #81, `fault-perception` in #83, `fault-pid-alert` in #78) is likely to duplicate the pattern since it was copy-pasted from the original UDS Makefile template.

## Proposal

Unify all per-module `fault-<module>` targets on the same exit-code-propagating template that PR #85 applies to `fault-uds`. Concretely, for each module:

```make
fault-<module>:
    @mkdir -p $(VV_REPORT_DIR_<module>)/fault_injection
    $(CC) ... -o $(VV_REPORT_DIR_<module>)/fault_injection/test_<module>_fault ...
    @$(VV_REPORT_DIR_<module>)/fault_injection/test_<module>_fault \
        > $(VV_REPORT_DIR_<module>)/fault_injection/run.log 2>&1; \
        rc=$$?; \
        cat $(VV_REPORT_DIR_<module>)/fault_injection/run.log; \
        echo ""; \
        if [ "$$rc" = "0" ]; then \
            echo "(all fault assertions PASS — flip continue-on-error to false in vv-<module>.yml)"; \
        else \
            echo "(non-zero exit expected while bugs are pending patch)"; \
        fi; \
        exit $$rc
```

Sequencing:
1. Apply to `fault-decision` on `development` directly (small PR, author-owned).
[...]

PR #83, opened 2026-04-23
test(vv-perception): cross-validation V&V stack for Perception module

jessica-leoa:
# Summary

- Adds independent cross-validation V&V stack for the Perception module (`aeb_perception.c` / `aeb_perception.h`) per ISO 26262-6:2018 §5.4.8 independence requirement
- Introduces 5 new Makefile targets for Perception V&V: `vv-perception`, `mcdc-perception`, `fault-perception`, `memory-perception`, `misra-perception`
- Adds `tests/test_perception_fault.c`: a fault-injection suite with 25 assertions organized in 4 categories (NaN/inf, numeric extremes, SEU/state corruption, persistence)
- Adds `reports/vv_perception/` with machine-generated evidence: MC/DC coverage logs, fault injection results, Valgrind/ASan/UBSan logs, cppcheck MISRA XML, functional test CSV and reproducibility README
- Reveals **3 robustness bugs** in the Perception module with proposed patches: (Bug 1 HIGH) NaN/∞ propagation through sensor fault detector, (Bug 2 MEDIUM) unsanitized `v_ego` passthrough, (Bug 3 MEDIUM) `prev_d` contamination causing false-positive fault latches
- Achieves 100% statement coverage and 88.04% branch/MC/DC — gap analysis for the 11 uncovered condition outcomes is documented with a closure plan
- Updates `.gitignore` to exclude V&V intermediate artefacts (binaries, `.gcda`, `.gcno`, `.tex`, `.pdf`) from the repository

# Related Issue

Related to the 3 robustness issues identified by this V&V activity, to be fixed on a `fix/perception-robustness` branch by the original author:

- Related to #79 — guard sensor fault detectors against NaN / ±Inf (Bug 1, HIGH)
- Related to #80 — sanitize `v_ego` passthrough and propagate fault on invalid input (Bug 2, MEDIUM)
- Related to #82 — update `prev_d` / `prev_vr` only on frames classified as good (Bug 3, MEDIUM)

# Change Type

- [ ] feat
- [ ] fix
- [ ] docs
- [x] test
- [ ] ci
- [x] refactor/chore

# AEB Areas Affected

- [x] Perception
- [ ] Decision Logic
- [ ] Driver Alerts
- [ ] Braking Control
- [ ] State Machine
- [ ] CAN Interface
- [ ] UDS Diagnostics
- [ ] Integration
- [ ] Documentation only

# Requirements Impacted

## Functional Requirements
- FR-PER-001 (distance output)
[...]

PR #81, opened 2026-04-23
test(can): ASIL-D V&V cross-validation of CAN interface

mphumapo2014:
Activities applied per ISO 26262-6:2018:
  - Table 10 item 1b  nominal tests          13 tests, 36 asserts, 100% PASS
  - Table 12 item 1c  MC/DC coverage         100% stmt/branch/MC/DC
  - Table 11 item 1e  fault injection        31 asserts, 31 PASS (4 categories)
  - Table  8 item 1d  memory safety          Valgrind 0, ASan 0, UBSan 1 UB
  - Table  8 item 1b  MISRA static scan      2 style advisory in-scope

Files added:
  - tests/test_can_struct.c   structural complementary tests (Lourenço)
  - tests/test_can_fault.c    systematic fault injection    (Lourenço)
  - reports/vv_can/           full evidence + consolidated report (generated locally)

Files modified:
  - Makefile: add mcdc-can, fault-can, memory-can, misra-can, vv-can targets

Files NOT modified (preserves author independence per ISO 26262-6 §5.4.8):
  - src/communication/aeb_can.{c,h}
  - stubs/can_hal.{c,h}
  - tests/test_can.c

Bug #1 found and documented: (uint32_t) cast in encode_unsigned is undefined
behaviour on NaN/Inf input (confirmed by UBSan). Patch proposed in the
consolidated report; will ship in a separate PR (fix/can-encode-robustness)
authored by Renato.

Reproducible: make vv-can
Full report: generated locally via `make vv-can` (not committed)

---

# Summary

- Independent cross-validation of the CAN interface module (`aeb_can.{c,h}`) following ISO 26262-6 §5.4.8 independence principle — same pipeline applied by Renato for UDS (`test/vv-uds`).
- Two complementary test suites added: `test_can_struct.c` (closes structural gaps in defensive switches) and `test_can_fault.c` (systematic fault injection across 4 categories: NULL / NaN / SEU / persistence).
- Consolidated structural coverage (nominal + structural + fault, merged via `gcov-tool`) achieves **100% statement, 100% branch, and 100% MC/DC** in `aeb_can.c`.
- Dynamic verification (Valgrind + ASan + UBSan) revealed **1 undefined behaviour** in `encode_unsigned` (float→uint cast on NaN/Inf input), documented as Bug #1 with a proposed fix; patch will be delivered in a separate PR to preserve independence.
- MISRA C:2012 static analysis scoped to `aeb_can.{c,h}`: 0 errors, 2 justified style advisories.
[...]

PR #78, opened 2026-04-23
tests(pid+alert): ASIL-D V&V cross-validation of PID and Alert modules

RianLinhares:
# Summary
- Cross-validation of the PID controller and Alert mapping modules (`aeb_pid.c` / `aeb_pid.h` and `aeb_alert.c` / `aeb_alert.h`) — original author: Jéssica Roberta de Souza Santos, cross-validator: Rian Ithalo da Costa Linhares, per §Cross-Assignment of `final_delivery_strategy.tex`.
- Adds `tests/test_pid_mcdc.c` and `tests/test_alert_mcdc.c` — MC/DC structural complementary tests covering 4 atomic defensive conditions that were uncovered by the nominal suite.
- Adds `tests/test_pid_fault.c` and `tests/test_alert_fault.c` — 22-test / 36-assertion systematic fault-injection suite across 4 categories (NaN/Inf, numeric extremes, SEU-simulated FSM state corruption, state persistence).
- Adds V&V Makefile targets (`mcdc`, `fault`, `memory`) for the Ubuntu 24.04 + gcc-14 + cppcheck + Valgrind + ASan/UBSan + gcovr toolchain.
- Ships reproducible evidence artefacts under `reports/vv_pid_alert/` (coverage, fault logs, memory-safety logs, MISRA XML, functional-test CSV).
- Existing CI gates (`make test`, `make misra`) remain green.

# Related Issue
Surfaces the evidence behind #68  (NaN propagation — HIGH) and #69  (fsm_state underconstrained — CRITICAL). Patches land in a separate `fix/pid-robustness` PR by the original author (Jéssica), per ISO 26262-6 §5.4.8 (independence between verification and implementation).

# Change Type
- [ ] feat
- [ ] fix
- [ ] docs
- [x] test
- [ ] ci
- [ ] refactor/chore

# AEB Areas Affected
- [ ] Perception
- [ ] Decision Logic
- [x] Driver Alerts
- [x] Braking Control
- [ ] State Machine
- [ ] CAN Interface
- [ ] UDS Diagnostics
- [ ] Integration
- [ ] Documentation only

# Requirements Impacted
## Functional Requirements
- FR-BRK-001 (progressive deceleration, no jumps > 2 m/s²)
- FR-BRK-002 (PI controller with anti-windup)
- FR-BRK-003 (capable of ≥ 5.0 m/s² braking demand)
- FR-BRK-004 (jerk limited to |jerk| ≤ 10 m/s³)
[...]

PR #91, opened 2026-04-24
ci(vv-decision): publish HTML reports to GitHub Pages

### LINKED ISSUE #87: ci(vv-decision): publish HTML reports to GitHub Pages
## Context

The Decision module (`src/decision/aeb_ttc.c` / `src/decision/aeb_fsm.c` and their headers) ships with a nominal unit test suite (`tests/test_decision.c`, 9 assertions) that runs in the existing CI (`make test`). PR #84 extends it with a 39-case MC/DC-complementary suite (`tests/test_decision_mcdc.c`) and a 32-assertion fault-injection suite (`tests/test_decision_fault.c`). For ASIL-D conformance per ISO 26262-6:2018, this module needs those structural and dynamic evidence sources wired into CI so they become enforced gates rather than ad-hoc local runs.

## Goal

Add a per-module CI workflow that runs the full ASIL-D V&V stack on the Decision module, gates the blocking metrics (MC/DC, memory safety, MISRA), and publishes navigable HTML reports for the assessor (banca) without requiring branch-protection bypass.

### V&V validation stack (in addition to the existing nominal + MCDC-complementary suites)

- **MC/DC coverage** (`make mcdc-decision`) --- gcc-14 `-fcondition-coverage` + gcov-14 + gcovr, gate ≥ 95% for statement, branch and MC/DC (current baseline measured in #84: 97.7% / 96.8% / 96.6%).
- **Memory safety** (`make memory-decision`) --- Valgrind + AddressSanitizer + UBSan on the nominal and MCDC-complementary binaries (gate: 0 errors), plus the same instrumentation on the fault binary (informational).
- **MISRA C:2012 conformance** (`make misra-decision`) --- cppcheck with the MISRA addon scoped to `aeb_ttc.{c,h}` / `aeb_fsm.{c,h}`, gate: 0 `error` / `warning` / `performance` / `portability` in module scope (the 12 `style` findings catalogued in #84 are documented deviations, not gate failures).
- **Fault injection** (`make fault-decision`, `tests/test_decision_fault.c`) --- 32 assertions across four categories (NaN/±Inf, numeric extremes, SEU-corrupted state, persistence); non-blocking until `fix/decision-robustness` lands and the four known robustness bugs (#64, #65, #66, #67) are patched.

### Report publication (reuses the infrastructure from #86)

- **Workflow Artifacts** on every run (PR or push), retention 30 days, available from the Actions tab of each individual run.
- **GitHub Pages** on push to `development` / `main` --- HTML reports under `https://erycaaf.github.io/AEB-stellantis-project/vv_decision/<branch>/`, deployed via `peaceiris/actions-gh-pages@v4` with `keep_files: true` so the Decision subtree coexists with the UDS subtree (and future modules) on the same `gh-pages` branch.
- **Landing page** at `docs/index.html` extended with a Decision section alongside the existing UDS cards.
- **HTML generators** reuse the scripts introduced in #86: `lcov genhtml` for coverage, `cppcheck-htmlreport` for MISRA, `scripts/wrap_memory.sh` for Valgrind/ASan/UBSan logs, `scripts/wrap_fault.py` for the fault-injection table.

### Why not commit-back

The commit-back pattern would require bypass of branch protection rules --- not exposed in classic Branch Protection (only Copilot Apps appear as selectable bypass actors) and would force migration to Rulesets. The Workflow Artifacts + GitHub Pages pattern is the industry-standard alternative and needs no bypass.

## Acceptance criteria

- [ ] `vv-decision.yml` runs on PRs and pushes to `development` / `main` with the four steps gated as described.
- [ ] Reports are uploaded as Workflow Artifacts on every run.
- [ ] On push to `development` / `main`, HTML reports appear under `https://erycaaf.github.io/AEB-stellantis-project/vv_decision/<branch>/`.
- [ ] Landing page at `https://erycaaf.github.io/AEB-stellantis-project/` lists Decision reports.
- [ ] No branch-protection bypass needed for `github-actions[bot]`.
- [ ] Reproducible locally via `make vv-decision && make html-decision` on Ubuntu 24.04 (or via `docker/Dockerfile.vv`).

## Out of scope

- The Decision V&V Makefile targets (`make mcdc-decision`, `fault-decision`, `memory-decision`, `misra-decision`).
- The shared CI infrastructure (`scripts/wrap_*`, `docs/index.html` skeleton, `publish-index.yml`, `publish-pages` pattern).
- Equivalent workflows for other modules (Perception, PID + Alert, CAN) --- those will follow the same pattern in dedicated PRs.
- Promoting `fault-decision` to blocking --- waits for `fix/decision-robustness` (which applies the patches for #64, #65, #66, #67).
- Per-path Rulesets configuration --- not needed in this approach.

erycaaf:
## Summary
- Add the full ASIL-D V&V workflow for the Decision module (`vv-decision.yml`): MC/DC coverage, memory safety (Valgrind + ASan + UBSan), MISRA C:2012, fault injection.
- Blocking gates (MC/DC, memory, MISRA) enforce ≥95% thresholds on PRs to `development`/`main`; fault injection is non-blocking until `fix/decision-robustness` lands.
- Add `make html-decision` target generating navigable HTML via `lcov genhtml`, `cppcheck-htmlreport`, and the module-agnostic wrapper scripts already in the repo.
- Unify `fault-decision` with the exit-code-propagating template so the CI step colour reflects reality once `continue-on-error` flips to blocking.
- Add `gcov --conditions` (`-g`) to `mcdc-decision` so `gcov_summary.txt` exposes the `Condition outcomes covered:` entries that the MC/DC gate parses.
- Add Decision V&V cards to `docs/index.html`, publishing the four artefact bundles per branch (`main`, `development`) under `vv_decision/<branch>/` on `gh-pages`.

## Related Issue
Closes #87
Addresses #88 for the `fault-decision` target on `development`; the remaining open V&V PRs (#78, #81, #83) are expected to adopt the same template before merge.

## Change Type
- [x] ci
- [x] build
- [x] docs

## AEB Areas Affected
- [x] Decision Logic

## Requirements Impacted
**Functional Requirements**
- FR-DEC-001 to FR-DEC-005 (V&V evidence: structural coverage, memory safety, MISRA, fault tolerance)

**Non-Functional Requirements**
- NFR-COD-001 (MISRA C:2012 conformance reporting)
- NFR-COD-006 (statement coverage gate ≥ 95%)
- NFR-COD-007 (branch / MC/DC coverage gate ≥ 95%)
- NFR-SAF-002 (fault tolerance evidence via fault injection)

## Artifacts Updated
- [x] CI workflow
- [x] Build system (Makefile)
- [x] README / SCM documentation (landing page)

## Validation Evidence
- [x] Local build passes
- [x] Automated tests pass
- [x] Traceability preserved (existing coverage_summary.txt + cppcheck XML artefacts intact)

## Evidence

The existing Decision unit suite (`tests/test_decision.c`, MC/DC suite `tests/test_decision_mcdc.c`, and fault-injection suite `tests/test_decision_fault.c`) covers functional behaviour and robustness. This PR adds the ASIL-D V&V evidence required by ISO 26262-6:2018 as an automated pipeline:

| Step | Tool | Gate | Source |
|---|---|---|---|
| MC/DC coverage | `gcc-14 -fcondition-coverage` + `gcov --conditions` | ≥ 95% (stmt / branch / MC/DC) per file | `make mcdc-decision` |
| Memory safety | Valgrind + ASan + UBSan | 0 errors on nominal + MC/DC suites | `make memory-decision` |
| MISRA C:2012 | `cppcheck` + MISRA addon | 0 issues in `aeb_ttc.{c,h}` / `aeb_fsm.{c,h}` | `make misra-decision` |
| Fault injection | `tests/test_decision_fault.c` | informational until `fix/decision-robustness` | `make fault-decision` |

## What the workflow does

- Triggers on changes to `src/decision/aeb_ttc.c`, `src/decision/aeb_fsm.c`, their headers, the three Decision test files, the Makefile, the shared wrappers, `docker/Dockerfile.vv`, or the workflow itself.
- Installs the V&V toolchain matching `docker/Dockerfile.vv` (gcc-14, cppcheck 2.13, Valgrind, lcov, gcovr). `cppcheck-htmlreport` is fetched from upstream pinned to tag `2.13.0` for reproducibility.
- MC/DC gate reads `gcov_summary.txt` and picks the worst-performing file across TTC and FSM (`sort -n | head -1`) so neither can regress silently behind the other.
- Memory-safety gate checks the four nominal/MC/DC logs; the fault-suite logs are tolerated as evidence until `fix/decision-robustness` merges.
- MISRA gate filters findings by `id.startswith("misra-c2012")` so cppcheck-native style findings do not trip the gate.
- On push to `development`/`main`, the `publish-pages` job deploys the artefact under `vv_decision/<branch>/` on `gh-pages` via `peaceiris/actions-gh-pages@v4` with `keep_files: true`, sharing the `gh-pages-deploy` concurrency group to avoid non-fast-forward races with other deploys.

## Resulting URLs

| Target | URL |
|---|---|
| Landing page | https://erycaaf.github.io/AEB-stellantis-project/ |
| Decision on `main` | https://erycaaf.github.io/AEB-stellantis-project/vv_decision/main/coverage_html/ |
| Decision on `development` | https://erycaaf.github.io/AEB-stellantis-project/vv_decision/development/coverage_html/ |

## Reproduction (local)

```bash
sudo apt-get install -y gcc-14 cppcheck python3-pygments valgrind lcov gcovr python3
sudo curl -fsSL https://raw.githubusercontent.com/danmar/cppcheck/2.13.0/htmlreport/cppcheck-htmlreport \
  -o /usr/local/bin/cppcheck-htmlreport && sudo chmod +x /usr/local/bin/cppcheck-htmlreport
make vv-decision     # runs MC/DC + memory + MISRA + fault + html
xdg-open reports/vv_decision/coverage_html/index.html
```

## Reviewer Notes

- The `fault-decision` rewrite implements the template proposed in #88: captures the binary's `rc`, picks a conditional message (PASS → "flip `continue-on-error` to false in `vv-decision.yml`"; FAIL → "non-zero exit expected while bugs are pending patch"), and re-exits with the real code. `continue-on-error: true` at the YAML level keeps the step non-blocking until `fix/decision-robustness` lands.
- `make html-decision` fails loudly with `exit 1` if `mcdc-decision` / `misra-decision` haven't produced their upstream artefacts, rather than publishing empty HTML bundles to `gh-pages`.
- MC/DC gate takes the worst-performing file across TTC and FSM — the gate only passes when **both** files meet ≥95% on statement, branch, and MC/DC.
- `--conditions` added to the two `gcov` calls in `mcdc-decision` so `gcov_summary.txt` contains the `Condition outcomes covered:` line. Does not change what is measured (already compiled with `-fcondition-coverage`), only what is displayed.
- Publish-pages uses the shared `gh-pages-deploy` concurrency group established by the UDS workflow, so parallel pushes from different module workflows serialise instead of racing.

## Risks / Open Points

- First post-merge push to `development` or `main` will create the `vv_decision/<branch>/` subtree on `gh-pages`. Subsequent runs add files alongside (no destructive replacement) thanks to `keep_files: true`.
- The fault-injection step stays non-blocking until `fix/decision-robustness` merges. After that, flip `continue-on-error` to `false` in `vv-decision.yml` and remove the conditional PASS message from the Makefile target.
- The `lcov --extract` patterns and the MISRA target tuple hardcode `aeb_ttc` / `aeb_fsm`. If the Decision module grows a third source file in the future, both need updating.
- The docs/index.html placeholder is updated manually as modules come online; automating the index generation is tracked in the team CI guide.


renatosfagundes (changes requested):
The CI failure is **your gate working exactly as designed** — that's the good news. Your `misra-c2012` filter is correctly scoped (copied cleanly from #85 after your own review note on item 3, nice) and it's correctly identifying that the Decision source has 12 MISRA-C:2012 findings. The bad news is that the gate is set to 0 and your own module has 12.

Breakdown of the 12 from `reports/vv_decision/misra/cppcheck_decision.xml`:

| Rule | Count | Classification | Meaning |
|---|---|---|---|
| **15.7** | 3 | **Required** | All `if...else if` constructs shall terminate with an `else` statement |
| **10.3** | 2 | **Required** | No essential-type narrowing in assignments |
| **15.5** | 5 | Advisory | A function should have a single exit point |
| **8.7**  | 2 | Advisory | Functions/objects with external linkage if referenced in only one TU |

So 5 **Required** and 7 **Advisory**. In industry practice, Advisory findings are typically handled by a documented deviation analysis (ISO 26262-8 §9.4.2) rather than a hard CI block — the 15.5 "single exit point" rule in particular is one most automotive projects deliberately waive.

Where in the code:

- `src/decision/aeb_fsm.c:190, 208` — Rule 10.3 (Required)
- `src/decision/aeb_fsm.c:168, 186, 204, 226` + `src/decision/aeb_ttc.c:62` — Rule 15.5 (Advisory)
- `src/decision/aeb_fsm.c:265, 313` + `src/decision/aeb_ttc.c:33` — Rule 15.7 (Required)
- `src/decision/aeb_ttc.c:13, 40` — Rule 8.7 (Advisory)

### Request — pick a path before this can merge

Three options, ordered by my preference:

**(A) Fix the 5 Required findings + scope the gate to Required/Mandatory only.** Rule 15.7 is usually a quick `else { /* no action */ }` addition; Rule 10.3 typically means an explicit `(uint8_t)` cast or declaring the target variable at the right width. Advisory rules stay in the code with a one-line deviation note in the consolidated V&V report §6 or similar. The gate code changes from "fail on any `misra-c2012`" to something like:

```python
# MISRA C:2012 Required + Mandatory rules (blocking).
# Advisory rules are informational and surface in the XML for the assessor
# but don't fail the gate — see report §6 for the deviation analysis.
ADVISORY = {
    "misra-c2012-8.7",
    "misra-c2012-15.5",
    # ... extend as the module set grows ...
}
for e in t.findall(".//error"):
    rid = e.get("id", "")
    if not rid.startswith("misra-c2012"):
        continue
    if rid in ADVISORY:
        continue
    loc = e.find("location")
    if loc is not None and loc.get("file", "") in target:
        n += 1
```

You might want to keep the advisory set in a repo-shared YAML/JSON so the UDS, PID+Alert, Perception, and CAN workflows all read from the same source once they catch up.

**(B) Fix all 12.** Cleanest result, bigger diff, more review cycles. If 15.5 isn't something you're going to refactor five functions to comply with, this option is effectively "rewrite the Decision module" — probably too heavy for this PR.

**(C) Keep the gate strict + document 12 as a known-failing baseline.** Not recommended. Once the gate is blocking with known-failing content, people start disabling it or marking failures "expected" — defeats the point of having the gate.

My recommendation is (A). That also establishes the advisory-policy precedent that the other open V&V PRs (#78, #83, #89) will almost certainly need when their module MISRA gates flip on (they'll hit 15.5 too — it's everywhere). If you agree, I'm happy to open a follow-up PR with the shared advisory list once this lands.

### Minor — step title should match the policy

`MISRA (0 issues in aeb_ttc/aeb_fsm)` will read confusingly once the policy accepts 7 advisory findings. Once you pick (A):

```yaml
- name: MISRA required/mandatory (0 findings in aeb_ttc/aeb_fsm)
```

Or similar. Not blocking; cosmetic.

---

Everything else in the PR looks clean — `html-decision` target, index cards, concurrency group on the publish job, fault-decision adopting the #88 template. Once the MISRA gate policy is decided, this is ready.

renatosfagundes:
Revising my earlier review — important context I had backwards.

`aeb_ttc.c` and `aeb_fsm.c` are Lourenço's code, not yours. This PR is the Decision V&V workflow, cross-validating his module. Per the author-independence principle (ISO 26262-6 §5.4.8) the project has been applying across every V&V PR (see UDS in #77, PID/Alert in #78, Perception in #83), **you shouldn't be the one patching those files** — my "fix the 5 Required violations" suggestion was pointing the ask at the wrong person.

The right pattern here is exactly the one I used for `fault-uds` in #85: the MISRA-Required gate stays in the workflow, but marked `continue-on-error: true` with a TODO pointing at a companion fix PR authored by Lourenço. When that fix merges and the source is clean, we flip the flag to blocking.

## Suggested change to `vv-decision.yml`

```yaml
# ─── Non-blocking until fix/decision-misra lands ──────────────────────
# The 5 MISRA Required findings in aeb_ttc.c / aeb_fsm.c are tracked
# in issue #<NEW> and will be patched by the original author (Lourenço)
# on fix/decision-misra. After that PR merges, flip continue-on-error
# to false (or remove) so this gate becomes blocking again. The 7
# Advisory findings (Rules 15.5, 8.7) are documented as deviations per
# ISO 26262-8 §9.4.2 in the consolidated V&V report §6.
- name: MISRA required/mandatory (0 findings in aeb_ttc/aeb_fsm)
  continue-on-error: true
  run: |
    make misra-decision
    DEC_ISSUES=$(python3 - <<'PY'
    import xml.etree.ElementTree as ET
    t = ET.parse("reports/vv_decision/misra/cppcheck_decision.xml")
    target = (
        "src/decision/aeb_ttc.c",
        "src/decision/aeb_fsm.c",
        "include/aeb_ttc.h",
        "include/aeb_fsm.h",
    )
    # MISRA C:2012 Advisory rules — informational, do not block.
    # Candidates for deviation analysis; not failure signals.
    ADVISORY = {
        "misra-c2012-8.7",
        "misra-c2012-15.5",
    }
    n = 0
    for e in t.findall(".//error"):
        rid = e.get("id", "")
        if not rid.startswith("misra-c2012"):
            continue
        if rid in ADVISORY:
            continue
        loc = e.find("location")
        if loc is not None and loc.get("file", "") in target:
            n += 1
    print(n)
    PY
    )
    echo "MISRA Required/Mandatory findings in Decision scope: $DEC_ISSUES"
    if [ "$DEC_ISSUES" != "0" ]; then
      echo "::error title=MISRA regression::$DEC_ISSUES Required/Mandatory finding(s) in aeb_ttc/aeb_fsm"
      exit 1
    fi
```

That achieves three things at once:

1. **Advisory findings (7×: 15.5 ×5, 8.7 ×2) stop failing the gate.** They're surfaced in the XML for the assessor; industry practice treats these as documented deviations, not blockers.
2. **Required findings (5×: 15.7 ×3, 10.3 ×2) still fail the gate**, but under `continue-on-error: true` so the CI step goes red without blocking merge — accurate signal without unblocking the workflow PR.
3. **When Lourenço's `fix/decision-misra` PR lands and the 5 Required findings are resolved**, you drop the `continue-on-error` line and the gate becomes blocking again. Same life-cycle flip I built into `fault-uds` for #70/#71/#72.

I'll open a tracking issue in a moment (similar to #70 for UDS Bug #1) listing the 5 Required findings with line numbers and rule descriptions, so Lourenço has a concrete task.

## On the `ADVISORY` set

I only know the classification of the four rules actually triggered in your run (`8.7`, `10.3`, `15.5`, `15.7`). Future modules will hit more rules — when they do, extending the `ADVISORY` set each time is tedious and error-prone. Worth factoring into a shared `scripts/misra_advisory.py` or a YAML lookup table that every per-module workflow reads from. Not for this PR; just flagging so I can raise it on the next one (#93?) or in #88's scope.

## On the gcovr frustration

Separate thread — my earlier comment still stands. Your `GCOVR ?= $(if $(wildcard venv/bin/gcovr),venv/bin/gcovr,gcovr)` pattern in the Makefile is the right one (both local-venv and CI-apt paths work). If the failure is in a specific workflow step, paste the `mcdc-decision` log line that's crashing and I can diagnose concretely — the usual suspects are runner version (must be `ubuntu-24.04` for gcovr 7.2+), `--gcov-executable='gcov-14'` passthrough, or `--root` pointing at the wrong directory after the shared-`.o` technique.

---

Net change from my previous review: **still `CHANGES_REQUESTED`, but the change is to the gate policy, not to the Decision source code.** Once the YAML tweak + tracking issue are in, this is approvable on its own without waiting for Lourenço's fix PR.

renatosfagundes:
Tracking issue opened: #92 — lists the 5 Required findings with line numbers and fix guidance, scoped as `fix/decision-misra` work for Lourenço. Once that lands you can drop the `continue-on-error: true` flag from the MISRA step here.

(The `<NEW>` placeholder in my previous comment's YAML snippet should read `#92`.)

erycaaf:
Thanks! Gate change applied in e2d561e, dropped the (@mphumapo2014 ) from the comment to keep public artefacts name-free, everything else is verbatim. Happy to add the tracking issue number once you open it.

On gcovr: already fixed in 3b1b13b (venv + gcovr>=8.0). The MC/DC gate passed on the run after that commit, current failure is MISRA, which this commit addresses.

renatosfagundes (approved):
LGTM 👍 — verified against `e2d561e`.

Three commits since my last review, all on-point:

**`e2d561e` — MISRA gate split.** Does exactly what the revised recommendation asked for:
- `ADVISORY = { "misra-c2012-8.7", "misra-c2012-15.5" }` filters the 7 advisory findings out of the gate.
- Required/Mandatory (the 5 findings — Rule 10.3 ×2, Rule 15.7 ×3) still surface as `::error` annotations.
- `continue-on-error: true` on the step with a lifecycle TODO in the comment block — same shape as `fault-decision` and my `fault-uds`.
- Once `fix/decision-misra` (tracked in #92) merges, drop the flag and the gate is blocking again.

**`3b1b13b` — gcovr ≥ 8.0 in a venv.** Correct fix for the Ubuntu 24.04 apt situation. `gcovr 7.2` in `--json-summary` mode omits `condition_percent`; the `python3 -m venv venv && venv/bin/pip install 'gcovr>=8.0'` approach works because the Makefile's `GCOVR ?= $(if $(wildcard venv/bin/gcovr), ...)` falls through to the venv binary when it exists. Industry-standard pattern for "apt version lags what the tool needs".

**`d3b85ad` — gate on gcovr aggregate, not per-binary gcov.** Good catch. `test_decision` covers aeb_ttc linear paths, `test_decision_mcdc` covers aeb_fsm condition pairs — each binary's per-source gcov is a partial picture; the union is what matters. `gcovr` with multi-directory input gives the real number (97.7% line / 96.8% branch).

### One nit (optional)

The MISRA step's comment says:

> The 5 MISRA Required findings in aeb_ttc.c / aeb_fsm.c are tracked in **a follow-up issue**

Worth replacing `a follow-up issue` with `#92` so a future reader reviewing this file standalone (e.g. an assessor) doesn't have to dig through the PR history to find it. Three-character edit. Not blocking.

### Pattern to carry forward

The `ADVISORY` set is a natural seed for a shared `scripts/misra_advisory.py` or YAML config once #78 (PID+Alert), #83 (Perception), and #89 (CAN) each gate MISRA. I raised that in my follow-up comment and it's fair game for a sweep in #88's scope or a new infra issue. Don't need to do it now.

Approving.

RianLinhares (approved):
LGTM

PR #97, opened 2026-04-24
feat(can-uds): integrate UDS over CAN (0x7DF/0x7E8)

### LINKED ISSUE #96: [FEAT] Wire UDS through aeb_can / aeb_core — service ReadDID / ClearDTC / RoutineControl over CAN at runtime
## Context

The UDS module (`src/communication/aeb_uds.c`) implements the full ISO 14229 server — SID 0x22 (ReadDataByIdentifier), SID 0x14 (ClearDiagnosticInformation), SID 0x31 (RoutineControl), DTC latching (C1001/C1004/C1006) — and is exercised by 23 nominal + 18 fault-injection assertions at unit level (PR #77).

However, the ECU at runtime **never actually services UDS requests over CAN**. `aeb_core_step()` only calls `uds_monitor_faults()` (step 9) to latch DTCs on perception faults. It does not:

- receive UDS_Request frames (CAN ID 0x7DF) on the RX path,
- call `uds_process_request()` with a valid request struct,
- emit UDS_Response frames (CAN ID 0x7E8) on the TX path.

The `aeb_can` module currently recognises only 0x080 (BrakeCmd TX), 0x100 (EgoVehicle RX), 0x101 (DriverInput RX), 0x120 (RadarTarget RX), 0x200 (FSMState TX), 0x300 (Alert TX) — UDS IDs are ignored by `can_rx_process()` because they're not in the dispatcher.

## Consequence

FR-UDS-005 ("UDS request/response via CAN within one cycle") is only partially satisfied: the request/response *struct handling* is correct and tested, but the *CAN transport* is absent. Without this integration, any client that sends a ReadDID or RoutineControl over CAN gets silence.

This also blocks the planned SIL "UDS Diagnostic" scenario on `feat/sil-gazebo-docker` (stopped ego, dashboard panel for live UDS queries / DTC inspection / AEB enable-disable), which cannot demonstrate anything until the ECU actually answers.

### `src/communication/aeb_can.h`

- Add CAN message IDs:
  ```c
  #define CAN_ID_UDS_REQUEST   (0x7DFU)
  #define CAN_ID_UDS_RESPONSE  (0x7E8U)
  ```
- Add DLC constants (4 bytes request, 8 bytes response, per `aeb_uds.h` struct sizes).
- Extend `can_rx_data_t` with a `uds_request_t uds_request;` plus a `uint8_t uds_request_pending;` flag cleared after consumption.
- Declare `int32_t can_tx_uds_response(const uds_response_t *resp);`.

### `src/communication/aeb_can.c`

- In `can_rx_process()`, add the 0x7DF case:
  - Decode the 4-byte payload into `uds_request_t` (SID / DID_high / DID_low / value).
  - Set `state->last_rx.uds_request_pending = 1U`.
- Implement `can_tx_uds_response()`:
  - Pack the 8-byte `uds_response_t` into a CAN frame.
  - `can_hal_send(CAN_ID_UDS_RESPONSE, frame, CAN_DLC_UDS_RESPONSE)`.

### `src/integration/aeb_core.c`

Insert a new step between step 9 (UDS fault monitoring) and step 10 (CAN TX):

```c
/* ── 10. UDS request service ──────────────────────────────────────── */
if (can_rx.uds_request_pending != 0U)
{
    uds_response_t uds_resp = {0};
    uds_process_request(&state->uds,
                        &can_rx.uds_request,
                        &uds_resp,
                        &state->fsm,
                        &state->pid,
                        &state->ttc);
    (void)can_tx_uds_response(&uds_resp);
    /* Clear the pending flag so the same request isn't serviced twice */
    /* (needs a helper on can_state_t since can_rx was a local copy).  */
}
```

The "clear the pending flag" step needs a small API addition to `aeb_can` — either `can_ack_uds_request(can_state_t *state)` or similar. Pick whichever reads cleaner.

### Unit tests (`tests/test_can.c`)

- Add three tests:
  - `test_rx_uds_request` — inject a 0x7DF frame, verify `uds_request` struct populated + `uds_request_pending = 1`.
  - `test_tx_uds_response` — call `can_tx_uds_response()` with a known `uds_response_t`, verify the HAL TX captured the expected 8 bytes.
  - `test_uds_request_ack` — verify the flag clears after ack.

### Unit tests (`tests/test_integration.c`)

- Add at least one integration test that drives a full request/response round-trip through `aeb_core_step`: inject a 0x7DF "ReadDID 0xF101 (FSM state)" RX, run one step, assert a 0x7E8 frame with response SID 0x62 + the current FSM state was queued by the HAL TX.

### Documentation

- Update `sil/aeb_gazebo/README.md` CAN message map if present.
- Update the consolidated SRS / requirements doc's FR-UDS-005 trace entry to reference this integration as the completion point.

## Acceptance

- `make test` passes (all existing 133+ assertions + 3–4 new CAN tests + 1 new integration test).
- `make vv-uds` still shows the same 100% MC/DC on `aeb_uds.c` (no source change there, just the callers).
- `make misra` / `make misra-uds` / `make misra-can` still clean — no new MISRA violations introduced.
- A manual end-to-end check: send a `0x7DF 22 F1 01` frame into the CAN RX path, observe `0x7E8 62 F1 01 01` come out (FSM state = STANDBY encoded as data byte 1).

## Not in scope (follow-up)

- CAN DBC file update for 0x7DF / 0x7E8 messages (separate PR, doc-only).
- SIL-side Python UDS client, FastAPI endpoints, dashboard panel, new `uds_diagnostic` scenario — lands on `feat/sil-gazebo-docker` once this PR is merged.
- UDS ISO-TP multi-frame support — out of scope; current DIDs + responses all fit in single 8-byte CAN frames.
- UDS security access (SID 0x27), session control (SID 0x10), tester present (SID 0x3E) — not required by current FR-UDS set.

## Why this lands now

The SIL "UDS Diagnostic" scenario requested for `feat/sil-gazebo-docker` is gated on this. Also, completing FR-UDS-005 at the integration level removes the asymmetry where every other functional CAN frame is serviced at runtime but UDS is only tested in-process.

## Dependency

Independent of #95. Both can proceed in parallel; they don't touch overlapping code.

renatosfagundes:
# Summary
- Wire UDS server into the AEB 10 ms core cycle so requests on CAN ID 0x7DF produce a response on 0x7E8 within the same cycle.
- Add `CAN_ID_UDS_REQUEST` / `CAN_ID_UDS_RESPONSE` filters and DBC layout decoding in `aeb_can`.
- Expose `uds_request_t` + `uds_request_pending` on `can_rx_data_t` plus two helpers: `can_tx_uds_response()`, `can_ack_uds_request()`.
- Insert step 10 in `aeb_core_step` (between UDS fault monitoring and CAN TX) that invokes `uds_process_request()` with live FSM/PID/TTC outputs, transmits the response, and clears the pending flag.
- Lift the TX capture struct into `stubs/can_hal_test.h` so tests can inspect transmitted frame bytes with a single definition of `tx_record_t`.
- Add 3 unit tests (`test_can`) and 1 end-to-end test (`test_integration`) covering decode, encode, ack, and full round-trip.
- Scope is the ISO 14229 functional broadcast (0x7DF); physical addressing on 0x7E0 is deferred.

# Related Issue
Closes #96

# Change Type
- [x] feat
- [ ] fix
- [ ] docs
- [ ] test
- [ ] ci
- [ ] refactor/chore

# AEB Areas Affected
- [ ] Perception
- [ ] Decision Logic
- [ ] Driver Alerts
- [ ] Braking Control
- [ ] State Machine
- [x] CAN Interface
- [x] UDS Diagnostics
- [x] Integration
- [ ] Documentation only

# Requirements Impacted
## Functional Requirements
- FR-UDS-005 — UDS request/response via CAN within one cycle (CAN path previously unwired; now complete end-to-end).
- FR-CAN-002 — RX filter list extended to include 0x7DF.
- FR-CAN-003 — DBC signal encoding applied to `UDS_Request` (0x7DF) and `UDS_Response` (0x7E8).

- NFR-PERF-001 — request is serviced inside the same 10 ms cycle; no extra context switches.
- NFR-POR-003 — module portability preserved (`can_hal_test.h` is test-only; production code path unchanged).

# Artifacts Updated
- [ ] Requirements document
- [ ] Modeling document
- [x] C source code
- [x] Tests
- [ ] CI workflow
- [ ] CHANGELOG

# Validation Evidence
- [x] Local build passes
- [x] Automated tests pass
- [ ] CI passes
- [ ] Traceability updated
- [ ] Documentation updated

## Evidence

Commands executed:
```
make test
```

Per-suite results (188/188 passed; up from 167/167 on development):
```
test_smoke:        14/14
test_can:          51/51  (+3 UDS: rx_uds_request, tx_uds_response, uds_request_ack)
test_perception:   25/25
test_decision:      9/9
test_pid:          19/19
test_alert:        27/27
test_uds:          23/23
test_integration:  20/20  (+1 UDS round-trip)
```

Build flags: `-Wall -Wextra -Wpedantic -std=c99 -O2` — zero warnings.

Affected scenarios: none of the CCRs / CCRm / CCRb scenarios change behavior. The new step 10 is a no-op when `uds_request_pending == 0`, which is the case for every existing integration scenario.

# Reviewer Notes

- Step ordering in `aeb_core_step` — UDS request service runs **after** fault monitoring (so DTC queries reflect the state the monitor just observed) and **before** CAN TX (so there is no intra-cycle lag on the response).
- `tx_record_t` was extracted from `stubs/can_hal.c` into `stubs/can_hal_test.h`. It is included only by the stub and by tests, so it does not leak into production builds.
- On-wire layout for UDS frames is a 1:1 byte map (`uds_request_t` = 4 bytes, `uds_response_t` = 8 bytes), matching `aeb_system.dbc`. No bit-packing helpers are needed.
- RX path sets `uds_request_pending = 1U` on a valid 0x7DF frame; the core cycle guarantees exactly one service-then-ack per request.

# Risks / Open Points

- **Scope**: only 0x7DF (functional) is filtered and decoded. 0x7E0 (physical, per-ECU) is deferred until a real tool requires it.
- **No queue**: if two 0x7DF requests arrive in the same 10 ms window, the second overwrites the first before it is serviced. This matches the SRS (one cycle = one request) but is worth flagging for a future stress test.
- **MISRA**: `make misra-decision` / `make misra-uds` are unaffected; a system-wide MISRA sweep for the new code is planned under Deliverable 4 (separate task, not in this PR).
- **Follow-ups**: the SIL "UDS Diagnostic" scenario (Python TCP UDS client + FastAPI endpoints + dashboard panel) lands on `feat/sil-gazebo-docker` after this PR merges.

### DIFF
--- a/include/aeb_can.h
+++ b/include/aeb_can.h
@@ -75,7 +75,27 @@ typedef struct
     uint8_t   aeb_enable;        /**< 0/1                                */
     uint8_t   driver_override;   /**< 0/1                                */
 
-    /* From UDS_Request (0x7DF) — FR-UDS-005 */
+    /* From UDS_Request (0x7DF) — FR-UDS-005
+     *
+     * Concurrency contract (IMPORTANT on Zephyr target):
+     *   can_rx_process()           writes uds_request_pending = 1 from the
+     *                              CAN RX callback context.
+     *   aeb_core_step()            reads-and-clears it from the 10 ms cycle
+     *                              thread.
+     *
+     *   On the host stub the two are called sequentially from the same
+     *   thread, so there is no race.  On the real Zephyr driver the RX
+     *   callback runs in a separate context; the driver integration layer
+     *   MUST serialise access — either by marshalling RX frames into a
+     *   message queue consumed by the core tick, by protecting this flag
+     *   with an atomic operation / brief interrupt disable around the
+     *   read-and-clear pair, or equivalent.
+     *
+     *   Without such synchronisation, a request arriving between the
+     *   read of uds_request_pending and the call to
+     *   can_clear_uds_request_pending() is silently dropped.  This is
+     *   not catchable at unit-test level because the host stub never
+     *   exercises the race — hence the documentation here. */
     uds_request_t uds_request;         /**< Decoded UDS request frame    */
     uint8_t       uds_request_pending; /**< 1 when a new request arrived;
                                             cleared by can_clear_uds_request_pending()
--- a/include/aeb_can.h
+++ b/include/aeb_can.h
@@ -78,7 +78,8 @@ typedef struct
     /* From UDS_Request (0x7DF) — FR-UDS-005 */
     uds_request_t uds_request;         /**< Decoded UDS request frame    */
     uint8_t       uds_request_pending; /**< 1 when a new request arrived;
-                                            cleared by can_ack_uds_request() */
+                                            cleared by can_clear_uds_request_pending()
+                                            after the response transmits. */
 
     /* Status */
     uint8_t   rx_timeout_flag;   /**< 1 if radar RX timed out            */
@@ -203,15 +204,21 @@ void can_get_rx_data(const can_state_t *state,
 int32_t can_tx_uds_response(const uds_response_t *resp);
 
 /**
- * @brief Acknowledge the pending UDS request.
+ * @brief Mark the current UDS request as serviced and clear the pending flag.
  *
- * Clears the uds_request_pending flag after aeb_core_step() has serviced
- * the request.  Must be called exactly once per request to prevent
- * double-processing on the next cycle.
+ * The caller (aeb_core_step) is responsible for deciding *when* to call
+ * this — typically only after can_tx_uds_response() has confirmed the
+ * response transmitted successfully. Clearing the flag on TX failure
+ * silently drops the response and forces the requester into a P2_server
+ * timeout (ISO 14229-2 §6.3) instead of a best-effort retry at the ECU
+ * level. See aeb_core.c step 10 for the retry policy.
+ *
+ * Must be called exactly once per successfully-serviced request to
+ * prevent double-processing on the next cycle.
  *
  * @param[in,out] state  Module state.
  */
-void can_ack_uds_request(can_state_t *state);
+void can_clear_uds_request_pending(can_state_t *state);
 
 /* ── Signal encode/decode helpers (FR-CAN-003) ───────────────────────── */
 
--- a/include/aeb_can.h
+++ b/include/aeb_can.h
@@ -14,15 +14,18 @@
 
 #include "aeb_types.h"
 #include "aeb_config.h"
+#include "aeb_uds.h"
 #include <stdint.h>
 
 /* ── CAN Message IDs (from aeb_system.dbc) ───────────────────────────── */
-#define CAN_ID_BRAKE_CMD      (0x080U)   /**< 128 — AEB_BrakeCmd  TX     */
-#define CAN_ID_EGO_VEHICLE    (0x100U)   /**< 256 — AEB_EgoVehicle RX    */
-#define CAN_ID_DRIVER_INPUT   (0x101U)   /**< 257 — AEB_DriverInput RX   */
-#define CAN_ID_RADAR_TARGET   (0x120U)   /**< 288 — AEB_RadarTarget RX   */
-#define CAN_ID_FSM_STATE      (0x200U)   /**< 512 — AEB_FSMState TX      */
-#define CAN_ID_ALERT          (0x300U)   /**< 768 — AEB_Alert TX         */
+#define CAN_ID_BRAKE_CMD      (0x080U)   /**< 128  — AEB_BrakeCmd   TX    */
+#define CAN_ID_EGO_VEHICLE    (0x100U)   /**< 256  — AEB_EgoVehicle RX    */
+#define CAN_ID_DRIVER_INPUT   (0x101U)   /**< 257  — AEB_DriverInput RX   */
+#define CAN_ID_RADAR_TARGET   (0x120U)   /**< 288  — AEB_RadarTarget RX   */
+#define CAN_ID_FSM_STATE      (0x200U)   /**< 512  — AEB_FSMState   TX    */
+#define CAN_ID_ALERT          (0x300U)   /**< 768  — AEB_Alert      TX    */
+#define CAN_ID_UDS_REQUEST    (0x7DFU)   /**< 2015 — UDS_Request    RX    */
+#define CAN_ID_UDS_RESPONSE   (0x7E8U)   /**< 2024 — UDS_Response   TX    */
 
 /* ── CAN Frame Lengths (bytes) ───────────────────────────────────────── */
 #define CAN_DLC_BRAKE_CMD     (4U)
@@ -31,6 +34,8 @@
 #define CAN_DLC_RADAR_TARGET  (8U)
 #define CAN_DLC_FSM_STATE     (4U)
 #define CAN_DLC_ALERT         (2U)
+#define CAN_DLC_UDS_REQUEST   (4U)
+#define CAN_DLC_UDS_RESPONSE  (8U)
 
 /* ── RX Timeout ──────────────────────────────────────────────────────── */
 #define CAN_RX_TIMEOUT_CYCLES (3U)  /**< 3 missed frames → fault        */
@@ -70,6 +75,11 @@ typedef struct
     uint8_t   aeb_enable;        /**< 0/1                                */
     uint8_t   driver_override;   /**< 0/1                                */
 
+    /* From UDS_Request (0x7DF) — FR-UDS-005 */
+    uds_request_t uds_request;         /**< Decoded UDS request frame    */
+    uint8_t       uds_request_pending; /**< 1 when a new request arrived;
+                                            cleared by can_ack_uds_request() */
+
     /* Status */
     uint8_t   rx_timeout_flag;   /**< 1 if radar RX timed out            */
 } can_rx_data_t;
@@ -103,8 +113,9 @@ int32_t can_init(can_state_t *state);
 /**
  * @brief Process a single received CAN frame (called from RX callback).
  *
- * Decodes EgoVehicle (0x100) or RadarTarget (0x120) per DBC layout.
- * Resets the RX miss counter on valid frame.
+ * Decodes EgoVehicle (0x100), RadarTarget (0x120), DriverInput (0x101)
+ * or UDS_Request (0x7DF) per DBC layout.  Resets the RX miss counter
+ * on valid radar frame.  A valid 0x7DF frame sets uds_request_pending.
  *
  * @param[in,out] state  Module state.
  * @param[in]     id     CAN message ID.
@@ -113,6 +124,7 @@ int32_t can_init(can_state_t *state);
  *
  * @req FR-CAN-002  Receive radar target data.
  * @req FR-CAN-003  DBC signal encoding.
+ * @req FR-UDS-005  Receive UDS request on 0x7DF.
  */
 void can_rx_process(can_state_t *state,
                     uint32_t     id,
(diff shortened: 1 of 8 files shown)


erycaaf (commented):
## Review — PR #97 · `feat/uds-can-integration`

Single-commit PR with clear scope: closes FR-UDS-005 by wiring UDS request/response on CAN 0x7DF/0x7E8 into the existing 10 ms cycle. The architectural move is right: pending flag on RX, service step inserted between UDS fault-monitor and CAN TX, 1:1 struct-to-wire mapping on encode/decode. Refactor of `tx_record_t` out of `can_hal.c` into `stubs/can_hal_test.h` is a good engineering move (replaces `extern` hack). Unit + integration coverage is present for the happy path. Scope note ("functional broadcast only, 0x7E0 physical addressing deferred") is explicit.

### Should Fix

**1. TX failure silently consumes the pending request** — `aeb_core.c:121-137`

```c
(void)can_tx_uds_response(&uds_resp);
can_ack_uds_request(&state->can);
```

If `can_hal_send()` fails (bus-off, TX mailbox full, HAL error), the response never goes out, but the pending flag is cleared unconditionally. The requester sees a silent timeout and must retry. Either:

- propagate the return and retain `uds_request_pending` on TX failure so the next cycle retries, or
- add a comment stating that retry is the requester's responsibility per ISO 14229-2 §6.3 (P2 timing) and acknowledging the trade-off.

For an ASIL-D context the current choice warrants an explicit rationale in-code, not just silence.

**2. Concurrency note for `uds_request_pending`** — `aeb_can.c` / `aeb_core.c`

`can_rx_process()` writes the flag; `aeb_core_step()` reads-and-clears it. In a real Zephyr deployment, `can_rx_process()` is called from the RX callback context, which is not the same thread as the 10 ms cycle. Without `volatile` or a critical section, a new request arriving between the read and `can_ack_uds_request()` is silently dropped.

Two acceptable paths:

- qualify the flag `volatile uint8_t` and briefly disable interrupts around the read-and-clear (or use an atomic op), or
- add a prominent comment on `can_rx_data_t` stating that the current stub runs single-threaded by construction and the real driver must serialise RX callback vs core tick.

The host-side stub never exercises this race, so unit tests will stay green either way; hence worth calling out explicitly.

**3. Naming of the ack helper** — `aeb_can.c:516-522`

`can_ack_uds_request()` reads like "UDS-level positive acknowledgement". What it actually does is "clear the pending flag regardless of response outcome". For audit readability I'd rename to `can_clear_uds_request_pending()` or `can_mark_uds_request_serviced()`. Cosmetic but matters when an assessor is scanning function names for intent.

**4. Negative-path coverage gap** — `tests/test_can.c`, `tests/test_integration.c`

Unit tests cover the happy path (valid 4-byte request, valid response TX, ack clears flag). Worth adding:

- short DLC (`dlc < CAN_DLC_UDS_REQUEST`): no `uds_request_pending` set, caller observes no-op. This is the only tested behaviour for malformed frames today — asserting it explicitly locks the contract.
- unknown SID or unknown DID at the UDS layer: confirm that `uds_process_request()` emits a negative response and that `can_tx_uds_response()` still transmits it. This is the ISO 14229 error-flow and it's worth a single-line assertion chain.

### What's Good

- **Requirement traceability throughout.** FR-UDS-005 cited in the commit body, the `@req` tags on `can_rx_process`, `can_tx_uds_response`, and the pipeline comment block in `aeb_core.h`.
- **Pipeline diagram in `aeb_core.h` kept in sync** with the inserted step 10 — documentation doesn't drift from code.
- **`uds_response_t uds_resp = {0};`** zero-init before `uds_process_request()` fills it — defensive against partial writes on error paths.
- **Refactor of `tx_record_t` into `stubs/can_hal_test.h`** replaces four `extern` declarations in `tests/test_can.c` with a clean header include. Also lets `tests/test_integration.c` inspect the TX capture for the round-trip test without duplicating the struct.
- **CAN IDs documented with decimal equivalents** in the header — practical for anyone reading a CAN sniffer trace.
- **`can_init()` registers the 0x7DF RX filter** alongside the existing filters — single add, no reorganisation.
- **Scope boundary explicit in the commit body** — "0x7DF only, 0x7E0 deferred" tells reviewers exactly what not to look for, which is rare and appreciated.
- **Integration test uses `find_tx(id)` to scan the capture buffer** rather than index-based assertions — robust to reordering of TX calls within a cycle.
- **Unit and integration tests route through `can_hal_test.h`** rather than replicating the struct — the refactor pays off immediately.
- **Null guards on all public APIs** (`can_ack_uds_request`, `can_tx_uds_response`) follow the existing module convention.

---

renatosfagundes:
Thanks for the review, @erycaaf. All 4 items addressed across three commits:

| # | Item | Fix | Commit |
|---|---|---|---|
| 1 | TX failure silently consumes the pending request | `aeb_core.c` now retains `uds_request_pending` when `can_tx_uds_response()` returns non-`CAN_OK`. The next 10 ms tick retries the transmission. Added an inline rationale citing ISO 14229-2 §6.3 P2_server timing — best-effort retry at the ECU layer is what ASIL-D wants for an idempotent DID read. | `05791c1` |
| 2 | Concurrency note for `uds_request_pending` | Documented the contract on the `can_rx_data_t` field: on the host stub both sides run sequentially in one thread, on the Zephyr target the RX callback and the 10 ms tick are separate contexts and the integration layer MUST serialise (message queue / atomic / brief IRQ disable). Chose documentation over a premature `volatile`, because `volatile` alone doesn't fix the read-and-clear race and would give a false sense of safety. | `c75046e` |
| 3 | Naming of the ack helper | Renamed `can_ack_uds_request` → `can_clear_uds_request_pending`. The former read like a UDS-level positive acknowledgement; the new name states what the function actually does. Docstring tightened to require `CAN_OK` on TX before calling. | `05791c1` |
| 4 | Negative-path coverage gap | Added three tests locking the malformed-frame and ISO 14229 error-flow contracts: `test_rx_uds_request_short_dlc` (DLC = 3 leaves pending at 0), `test_uds_negative_response_unknown_sid` (SID 0xAA → 0x7F + NRC 0x11), `test_uds_negative_response_unknown_did` (SID 0x22 with DID 0xFFFF → 0x7F + NRC 0x31). All three exercise the full CAN → UDS → CAN pipeline, not just the UDS layer in isolation. | `81ac4ea` |

Commits #1 and #3 landed together because the new function name encodes the new retry contract (the docstring on `can_clear_uds_request_pending` states "only call after `CAN_OK`"), so splitting them would have forced an inconsistent intermediate state.

### Test count

```
test_smoke:        14/14
test_can:          52/52  (+1 short-DLC)
test_perception:   25/25
test_decision:      9/9
test_pid:          19/19
test_alert:        27/27
test_uds:          23/23
test_integration:  26/26  (+2 negative-path)
---
total:            195/195  (up from 188/188)
```

Zero warnings under `-Wall -Wextra -Wpedantic -std=c99 -O2`.

Ready for another look.

erycaaf (approved):
## Re-review — PR #97

All four Should Fix items from the previous review are addressed across three focused commits. Verified each against the diffs:

| # | Item | Commit | Status |
|---|---|---|---|
| 1 | TX failure / silent ack (response dropped without retry) | `05791c1` | OK — `can_clear_uds_request_pending()` is now gated behind `can_tx_uds_response(...) == CAN_OK` in `aeb_core_step`. On TX failure, `uds_request_pending` stays 1 and the next 10 ms tick retries. Inline policy comment cites ISO 14229-2 §6.3 (P2_server timeout) as the alternative being avoided. |
| 2 | Concurrency contract on `uds_request_pending` | `c75046e` | OK — Explicit contract block on `can_rx_data_t.uds_request_pending` covering host-stub behaviour (serial on one thread), Zephyr integrator obligations (message queue, atomic op, or brief IRQ disable around the read-and-clear pair), and the rationale for not using `volatile` in isolation. Documented exactly what's not catchable at unit-test level. |
| 3 | Naming of the ack helper | `05791c1` | OK — Renamed `can_ack_uds_request` → `can_clear_uds_request_pending`. Docstring now states the call-after-CAN_OK contract. All call sites (`aeb_core.c`, `test_can.c`) updated. |
| 4 | Negative-path coverage | `81ac4ea` | OK — Three new tests: `test_rx_uds_request_short_dlc` (3-byte frame → `uds_request_pending` stays 0), `test_uds_negative_response_unknown_sid` (SID 0xAA → NRC 0x11 Service Not Supported), `test_uds_negative_response_unknown_did` (DID 0xFFFF → NRC 0x31 Request Out Of Range). Full pipeline end-to-end, confirming negative responses propagate on the wire instead of being dropped. |

Nice touch on 1: retry-on-failure is the right ASIL-D policy for idempotent DID reads / short RoutineControl requests, and the in-code comment makes the choice legible for the assessor.

Approving.

RianLinhares (commented):
LGTM

RianLinhares (approved):
LGTM

PR #94, opened 2026-04-24
fix(uds): address ASIL-D robustness gaps found in cross-validation

RianLinhares:
# Summary
- Apply three defensive patches in `src/communication/aeb_uds.c` identified by independent ASIL-D cross-validation (Relatório Consolidado V&V UDS, 21/04/2026).
- Bug 1 (HIGH): validate float inputs before `uint16_t` cast in `ReadDataByIdentifier` for DIDs `0xF100` (TTC) and `0xF102` (Brake), returning NRC `0x31` on NaN/±Inf/out-of-range.
- Bug 2 (CRITICAL, SEU): normalise `aeb_enabled` to strict Boolean in `uds_get_output` to defend against RAM bit-flips.
- Bug 3 (CRITICAL, SEU): normalise each DTC flag before summing `dtc_count` to prevent `uint8_t` wrap when flags are corrupted.
- Fixes converts fail-safe-by-coincidence (GCC 14 / x86-64) into fail-safe-by-design, independent of toolchain and optimisation level.
- No behavioural change for legitimate inputs: nominal suite (23/23) remains green.

## Related Issue
- Closes #70 
- Closes #71 
- Closes #72 

# Change Type
- [ ] feat
- [x] fix
- [ ] docs
- [ ] test
- [ ] ci
- [ ] refactor/chore

# AEB Areas Affected
- [ ] Perception
- [ ] Decision Logic
- [ ] Driver Alerts
- [ ] Braking Control
- [ ] State Machine
- [ ] CAN Interface
- [x] UDS Diagnostics
- [ ] Integration
- [ ] Documentation only

# Requirements Impacted
## Functional Requirements
- FR-UDS-001 (ReadDataByIdentifier — robustness on input domain)
- FR-UDS-002 (DTC storage — Boolean integrity of flags)

## Non-Functional Requirements
- NFR-SAF-ROB (robustness against invalid inputs and SEU)

# Artifacts Updated
- [ ] Requirements document
- [ ] Modeling document
- [ ] Simulink / Stateflow model
- [x] C source code
- [ ] Tests
- [ ] CI workflow
- [ ] README / SCM documentation
- [ ] CHANGELOG

# Validation Evidence
- [x] Local build passes
- [x] Automated tests pass
- [ ] CI passes
- [ ] Traceability updated
- [x] Safety-relevant impact assessed
- [ ] Documentation updated

## Evidence
[...]

PR #89, opened 2026-04-24
test(can): ASIL-D V&V cross-validation of CAN interface

mphumapo2014:
Activities applied per ISO 26262-6:2018:
  - Table 10 item 1b  nominal tests          13 tests, 36 asserts, 100% PASS
  - Table 12 item 1c  MC/DC coverage         100% stmt/branch/MC/DC
  - Table 11 item 1e  fault injection        31 asserts, 31 PASS (4 categories)
  - Table  8 item 1d  memory safety          Valgrind 0, ASan 0, UBSan 1 UB
  - Table  8 item 1b  MISRA static scan      2 style advisory in-scope

Files added:
  - tests/test_can_struct.c   structural complementary tests (Lourenço)
  - tests/test_can_fault.c    systematic fault injection    (Lourenço)
  - reports/vv_can/           full evidence + consolidated report (generated locally)

Files modified:
  - Makefile: add mcdc-can, fault-can, memory-can, misra-can, vv-can targets

Files NOT modified (preserves author independence per ISO 26262-6 §5.4.8):
  - src/communication/aeb_can.{c,h}
  - stubs/can_hal.{c,h}
  - tests/test_can.c

Bug 1 found and documented: (uint32_t) cast in encode_unsigned is undefined
behaviour on NaN/Inf input (confirmed by UBSan). Patch proposed in the
consolidated report; will ship in a separate PR (fix/can-encode-robustness)
authored by Renato.

Reproducible: make vv-can
Full report: generated locally via `make vv-can` (not committed)

---

# Summary

- Independent cross-validation of the CAN interface module (`aeb_can.{c,h}`) following ISO 26262-6 §5.4.8 independence principle — same pipeline applied by Renato for UDS (`test/vv-uds`).
- Two complementary test suites added: `test_can_struct.c` (closes structural gaps in defensive switches) and `test_can_fault.c` (systematic fault injection across 4 categories: NULL / NaN / SEU / persistence).
- Consolidated structural coverage (nominal + structural + fault, merged via `gcov-tool`) achieves **100% statement, 100% branch, and 100% MC/DC** in `aeb_can.c`.
- Dynamic verification (Valgrind + ASan + UBSan) revealed **1 undefined behaviour** in `encode_unsigned` (float→uint cast on NaN/Inf input), documented as Bug #1 with a proposed fix; patch will be delivered in a separate PR to preserve independence.
- MISRA C:2012 static analysis scoped to `aeb_can.{c,h}`: 0 errors, 2 justified style advisories.
[...]

PR #100, opened 2026-04-24
fix(pid): add robustness guards for NaN inputs and invalid FSM states

### LINKED ISSUE #68: [FIX] PID: guard pid_brake_step against non-finite a_ego / decel_target
## Sprint
Sprint 4 – Tests & Validation

## Due Date
2026-04-24

## Objective
`pid_brake_step` in `src/execution/aeb_pid.c` does not gate non-finite inputs. When `a_ego` or `decel_target` is `NaN`, the value propagates through the PI arithmetic chain (starting at the error term on line 138) and appears at the output as `brake_pct = NaN`. The `clamp_f32` output clamp cannot detect `NaN` because `NaN < 0` and `NaN > 100` are both false in IEEE 754. A brake actuator receiving `NaN` may interpret it as zero (no braking), as the maximum (wheels locked), or enter an error state — none of which is predictable. This defeats ASIL-B determinism.

## Scope
- Add a non-finite guard right after the existing NULL check (around `aeb_pid.c:102`):
  ```c
  if (!isfinite(decel_target) || !isfinite(a_ego))
  {
      /* Fail-safe: no braking, reset PI state */
      s_integral       = 0.0F;
      s_prev_brake_pct = 0.0F;
      output->brake_pct = 0.0F;
      output->brake_bar = 0.0F;
      return;
  }
Include <math.h> for isfinite.

## Acceptance Criteria
* Fault-injection assertions FAULT-A1 and FAULT-A4 (3 assertions) pass on test_pid_fault.c.
* All 12 nominal tests in test_pid.c and the complementary MC/DC tests continue to pass.
* make vv-pid (or make fault / make mcdc / make memory) re-runs clean.
* A PR is opened on fix/pid-robustness (base development) and linked to this issue.

## Related Requirements / Modules
* src/execution/aeb_pid.c:102-138 (pid_brake_step)
* NFR-SAF-ROB (robustness to invalid inputs)
* IEEE 754-2019 NaN comparison semantics

## Notes
Severity: HIGH. Documented in §7.1 of the Consolidated V&V Report for the PID/Alert module.

### LINKED ISSUE #69: [FIX] PID: use explicit whitelist of braking states in pid_brake_step guard
## Sprint
Sprint 4 – Tests & Validation

## Due Date
2026-04-24

## Objective
The braking-state guard at `aeb_pid.c:119` is written as:
```c
if ((fsm_state < FSM_BRAKE_L1) || (decel_target <= 0.0F))
```
Because FSM_BRAKE_L1 = 3, the test accepts any value >= 3 as a valid braking state — including out-of-enum values produced by an SEU (e.g. 7, 42, 99, 255). In that case the PI controller computes a braking command from a corrupted decel_target, potentially issuing a spurious emergency brake on an otherwise clear road. ASIL-D requires explicit protection against this class of single-event upset.

## Scope
Replace the implicit-blacklist guard with an explicit whitelist of the four valid braking states:
```c
const uint8_t is_braking_state =
    ((fsm_state == (uint8_t)FSM_BRAKE_L1) ||
     (fsm_state == (uint8_t)FSM_BRAKE_L2) ||
     (fsm_state == (uint8_t)FSM_BRAKE_L3) ||
     (fsm_state == (uint8_t)FSM_POST_BRAKE));

if ((!is_braking_state) || (decel_target <= 0.0F))
{
    /* reset + zero output (fail-safe) */
    ...
}
```
Any out-of-enum fsm_state now falls automatically into the fail-safe branch.

## Acceptance Criteria
* Fault-injection assertions FAULT-C1 and FAULT-C2 (3 assertions) pass.
* All 12 nominal tests in test_pid.c continue to pass — the whitelist covers exactly the four valid braking states.
* A PR is opened on fix/pid-robustness (base development) and linked to this issue.

## Related Requirements / Modules
* src/execution/aeb_pid.c:119 (braking-state guard)
* ASIL-D fail-safe principle; SEU protection
* FR-FSM-001 (valid enum), FR-BRK-002 (PI controller)

## Notes
* Severity: CRITICAL (SEU). Documented in §7.2 of the Consolidated V&V Report for the PID/Alert module.

jessica-leoa:
Addresses two ASIL-D robustness vulnerabilities identified during independent V&V cross-validation (see V&V Consolidated Report, Sections 7.1 and 7.2):

- Bug 1 (HIGH): NaN propagation through arithmetic chain. IEEE 754 NaN bypasses clamp_f32 because NaN comparisons return false. Added isfinite() guard for decel_target and a_ego with fail-safe output (zero brake) and integrator reset.
- Bug 2 (CRITICAL): Invalid fsm_state values (e.g., from Single-Event Upsets in RAM) interpreted as valid braking states by the implicit blacklist guard (fsm_state < FSM_BRAKE_L1). Replaced with explicit whitelist of the four valid braking states (L1, L2, L3, POST_BRAKE).

Guards unified into a single if preserving MISRA C:2012 Rule 15.5 (single exit point). State reset preserved on both s_integral and s_prev_brake_pct to avoid jerk limiter wind-up on fail-safe exit.

Validation (Codespace, Ubuntu, gcc-14 14.2.0):
- make test:              no regression (all nominal suites pass)
- make fault-pid-alert:   PID 18/18 PASS (was 12/18); Alert 18/18 (unchanged)
- make memory-pid-alert:  0 errors (Valgrind + ASan + UBSan)
- make mcdc-pid-alert:    new lines 100% covered (85/85 aggregate); condition coverage 90.6% (29/32) - down from baseline 100% (23/23) because the unified guard introduced 9 new atomic conditions (4x whitelist comparisons + 2x isfinite + 3x OR combinations). This is not a behavioral regression: it reflects the metric honestly responding to additional decision points. Complementary MC/DC test cases in tests/test_pid_mcdc.c to restore 100% are out of scope for this robustness fix and belong to the V&V author's next iteration.

Bugs identified by Rian Ithalo da Costa Linhares during independent V&V cross-validation. Post-patch re-validation pending on merged branch, per ISO 26262-6 §5.4.8 principle of independence.

Refs: ISO 26262-6:2018 Tables 8, 11; NFR-SAF-ROB

# Summary

- Adds `#include <math.h>` to `src/execution/aeb_pid.c` for `isfinite()` support.
- Replaces the implicit blacklist guard (`fsm_state < FSM_BRAKE_L1`) with an explicit whitelist of the four valid braking states (BRAKE_L1, BRAKE_L2, BRAKE_L3, POST_BRAKE).
- Adds explicit `isfinite()` rejection for `decel_target` and `a_ego` to block NaN/Inf propagation through the PI arithmetic chain.
- Unifies both new guards with the pre-existing `decel_target <= 0.0F` check into a single conditional, preserving MISRA C:2012 Rule 15.5 (single exit point).
- Preserves the dual state reset (`s_integral` and `s_prev_brake_pct`) in the fail-safe branch to prevent jerk limiter wind-up on exit.
- No changes to the PI control law, jerk limiter, output clamping, or `aeb_alert.c` (the Alert module already passed all fault injection tests).

# Related Issue

Closes #68
Closes #69

# Change Type

- [ ] feat
- [x] fix
- [ ] docs
- [ ] test
- [ ] ci
- [ ] refactor/chore

# AEB Areas Affected

- [ ] Perception
- [ ] Decision Logic
- [ ] Driver Alerts
- [x] Braking Control
- [ ] State Machine
- [ ] CAN Interface
- [ ] UDS Diagnostics
- [ ] Integration
- [ ] Documentation only

# Requirements Impacted

## Functional Requirements

- FR-BRK-001 to FR-BRK-007 (PID controller — behavior preserved, no functional change)

- NFR-SAF-ROB (robustness to invalid inputs — **now fully satisfied**, was partially violated per V&V Report §7)

# Artifacts Updated

- [ ] Requirements document
- [ ] Modeling document
- [x] C source code
- [ ] Tests
- [ ] CI workflow
- [ ] CHANGELOG

# Validation Evidence

- [x] Local build passes
- [x] Automated tests pass
- [ ] CI passes *(pending — will be verified by GitHub Actions on this PR)*
- [ ] Traceability updated *(out of scope — trace matrix update belongs to the V&V author's next iteration)*
- [ ] Documentation updated *(code-level comments expanded; external docs unchanged)*

## Evidence

**Environment:** GitHub Codespace, Ubuntu 24.04, gcc-14 (Ubuntu 14.2.0-4ubuntu2~24.04.1) 14.2.0, gcovr 8.6.

**Commands executed and results:**

```
$ make test
=== All test suites passed ===
(test_smoke 14/14, test_can 36/36, test_perception 25/25, test_decision 9/9,
 test_pid 19/19, test_alert 27/27, test_uds 23/23, test_integration 14/14)

$ make fault-pid-alert
=== PID Fault Injection Test Suite ===
Results: 18/18 passed, 0 failed       (baseline: 12/18 — 6 asserts FAIL→PASS:
                                       FAULT-A1 x2, FAULT-A4, FAULT-C1 x2, FAULT-C2)
=== Alert Fault Injection Test Suite ===
Results: 18/18 passed, 0 failed       (unchanged from baseline)

$ make memory-pid-alert
Valgrind   : 0 errors (both test_pid and test_alert)
ASan+UBSan : (clean) on both binaries

$ make mcdc-pid-alert
Statement coverage (aeb_pid.c, aggregate): 85/85 = 100%
Branch coverage (aeb_pid.c, aggregate):    34/37 = 91.9%
Condition coverage (aeb_pid.c, aggregate): 29/32 = 90.6%
aeb_alert.c: identical to baseline (patch does not touch this module)
```

**Affected scenarios:** this fix does **not change** nominal CCRs/CCRm/CCRb behavior — the whitelist covers exactly the four braking states that were already functionally active (L1, L2, L3, POST_BRAKE) under the previous blacklist. Behavioral change is limited to the fail-safe branch, which now additionally catches: (a) NaN/Inf inputs from corrupted sensors or upstream division-by-zero, (b) corrupted `fsm_state` values from Single-Event Upsets in RAM.

**Back-to-back impact:** none observed. All 19 nominal PID unit tests, all 14 integration tests, and all existing fault injection assertions (including FAULT-D1 integrator-reset-after-invalid-state-excursion and FAULT-D2 jerk-limit-under-spike) pass with the patch applied.

# Reviewer Notes

Three points deserve focused review:

1. **Whitelist completeness.** Confirm that `FSM_BRAKE_L1`, `FSM_BRAKE_L2`, `FSM_BRAKE_L3`, `FSM_POST_BRAKE` are the only FSM states in which the PID controller should be active. Any future addition of a new braking state will require updating this whitelist — consider whether a central definition (e.g. `IS_BRAKING_STATE(x)` macro in `aeb_types.h`) would be preferable.

2. **MC/DC coverage drop.** The aggregate condition coverage went from 100% (23/23) to 90.6% (29/32). This is a denominator effect: the unified guard introduces 9 new atomic condition outcomes (4× whitelist equality checks, 2× `isfinite()` calls, plus the combinatorial OR outcomes). Restoring 100% requires complementary cases in `tests/test_pid_mcdc.c` — which is out of scope here by task charter (testing infrastructure is the V&V author's responsibility).

3. **Style consistency.** The new `const uint8_t ... ? 1U : 0U` predicate pattern and explicit `== 0U` / `!= 0` comparisons follow the existing style in `src/perception/aeb_perception.c:146-147` and `src/execution/aeb_alert.c:146-161`. No new type (`bool_t` or `<stdbool.h>`) was introduced.

# Risks / Open Points

- **Post-patch V&V re-validation pending.** Per ISO 26262-6 §5.4.8, independent re-validation after a cross-validation fix belongs to the V&V author (Rian Ithalo da Costa Linhares). Expected after merge into `development`.
- **MC/DC test complement pending.** `tests/test_pid_mcdc.c` needs new cases to exercise each whitelist branch (BRAKE_L1, L2, L3, POST_BRAKE) independently and each `isfinite()` rejection path. Out of scope for this PR.
- **Traceability matrix update pending.** The requirement trace (NFR-SAF-ROB → code → test) should be updated to point to this commit as the robustness implementation.
- **No behavioral assumption change.** This fix preserves existing behavior for all nominal inputs; it only tightens the fail-safe branch. No upstream module (FSM, perception) needs to be adjusted.


renatosfagundes (changes requested):
## Review -- PR #100 · `fix/pid-robustness`

Clean fix addressing the two robustness gaps that Rian's cross-validation surfaced. The unified single-`if` preserves MISRA Rule 15.5 and reads well, and the whitelist approach for `fsm_state` is the right call for SEU protection -- the previous `fsm_state < FSM_BRAKE_L1` interprets any above-range corrupted value (e.g. FSM_OFF bit-flipped to a braking enum with the high bit set) as a valid braking state, which the whitelist categorically rejects.

One blocker on the PR metadata before this can merge.

### Must Fix

**PR body `Closes #` is empty -- issues won't auto-close**

The body references Bug #1 (HIGH) and Bug #2 (CRITICAL), but the `Related Issue` section has `Closes #` with no number. The whole reason these tracked issues exist is so a merge closes them automatically and the audit trail stays accurate. The two issues on the tracker are:

- #68 -- PID: guard `pid_brake_step` against non-finite `a_ego` / `decel_target`
- #69 -- PID: use explicit whitelist of braking states in `pid_brake_step` guard

Update the body to:

```
# Related Issue
Closes #68
Closes #69
```

GitHub will then auto-close both on merge.

### What's Good

- **Whitelist over blacklist.** `fsm_state == L1 || L2 || L3 || POST_BRAKE` is semantically correct for SEU resilience -- the previous `fsm_state < FSM_BRAKE_L1` approach interprets corrupted out-of-range values as "not braking", which happens to be fail-safe but is the right answer by coincidence, not by design.
- **Unified single-`if` preserves MISRA 15.5.** Combining `fsm_valid`, `inputs_finite`, and the existing `decel_target <= 0.0F` check into one conditional keeps the single-exit-point pattern consistent with the rest of `aeb_pid.c`.
- **`isfinite()` guards on both `decel_target` and `a_ego`.** Catches the whole arithmetic chain in one place rather than relying on `clamp_f32` (which silently passes NaN through because NaN comparisons are all false).
- **State reset preserved on the fail-safe branch.** `s_integral` and `s_prev_brake_pct` both get cleared, preventing integrator wind-up leaking into the next cycle once the fault clears.
- **MC/DC regression (100% -> 90.6%) honestly disclosed and deferred** to the V&V author's next iteration. The right call -- patching the code first, then adding complementary MC/DC tests on the V&V branch preserves the ISO 26262-6 §5.4.8 independence principle.
- **Scope discipline.** No changes to the PI control law, jerk limiter, output clamping, or `aeb_alert.c`. Pure defensive hardening.

jessica-leoa:
Thanks for the thorough review, @renatosfagundes ! Updated the PR body with
`Closes #68` and `Closes #69` — both will auto-close on merge.

Also appreciate the explicit callout on the blacklist-vs-whitelist
semantics — you're right that the previous `fsm_state < FSM_BRAKE_L1`
was fail-safe by coincidence rather than design. The whitelist makes
the intent explicit.

Ready for another pass when you have a moment.

renatosfagundes (dismissed):
pbpm

erycaaf (commented):
## Review — PR #100 · `fix/pid-robustness`

Two well-scoped robustness guards on `pid_brake_step` — `isfinite()` on arithmetic inputs plus a whitelist on `fsm_state`. The unified single-`if` preserves MISRA 15.5 and the fail-safe branch clears both `s_integral` and `s_prev_brake_pct` (no wind-up leaking into the next cycle). Coverage regression (100% → 90.6% MC/DC) is honestly disclosed and correctly deferred to the V&V author per ISO 26262-6 §5.4.8. Agreeing with the prior review's code-level assessment.

### Should Fix

**1. Branch is significantly behind `development` (31 files, 3200 lines missing).**

Staleness check:

```
$ git merge-base origin/development origin/fix/pid-robustness
8f59b65  tests(pid+alert): ASIL-D V&V cross-validation of PID and Alert modules  (2026-04-23 23:08)

$ git diff --stat origin/fix/pid-robustness..origin/development
...
31 files changed, 3200 insertions(+), 145 deletions(-)
```

Since the merge-base, `development` has absorbed the Decision CI pipeline (#91), Perception V&V (#96, `tests/test_perception_fault.c` + `test_perception_mcdc.c`), CAN over UDS (#97), and the `fault-uds` promote-to-blocking flip (#103). None of those touch `src/execution/aeb_pid.c` — the merge will be clean, no conflict expected — but the CI on this branch is running against a stale base. A `git merge origin/development` (or rebase) + push exercises the fix against the current test surface, which is the right pre-merge check for ASIL-D.

Same pattern as flagged on PR #89. Scope is much lower-risk here because the edit is single-file disjoint, so I'm marking this Should rather than Must — but worth doing before merge.

**2. Commit message names the V&V author in prose.**

`a77c4e3` commit body line:

> *"Bugs identified by Rian Ithalo da Costa Linhares during independent V&V cross-validation."*

Per the project convention (`@mentions` and handles, not prose names in commits / code / README), this should read either:

- "Bugs identified by `@RianLinhares` during independent V&V cross-validation" (keeps attribution, uses handle), or
- "Bugs identified during independent V&V cross-validation (see PR #XX / V&V Consolidated Report §7)" (neutral, cites artefact).

Fixable via `git commit --amend` + `git push --force-with-lease` if iteration is still open on the branch. If not, note for the next fix PR.

**3. Inline comment block duplicates the commit body.**

The comment on `pid_brake_step` grew from 7 lines to ~16 lines, adding paragraph-level explanations of IEEE 754 NaN semantics, SEU modelling, and the whitelist-vs-blacklist choice. The commit body already contains the same material with the same depth — two copies of the same rationale is a drift trap.

Defensible as documenting the non-obvious IEEE 754 motivation at the line — that's the one piece that genuinely helps a future maintainer avoid re-introducing the bug. Suggested collapse:

```c
/* Guard: controller active only in {L1, L2, L3, POST_BRAKE} with finite inputs.
 * Whitelist (not `< FSM_BRAKE_L1`) blocks SEU-corrupted out-of-range values;
 * isfinite() blocks NaN, which bypasses clamp_f32 via IEEE 754 compare-false.
 * Reset integrator to avoid wind-up across the fail-safe branch. */
```

Four lines carrying the non-obvious bits; the full derivation stays in the commit and the V&V report. Non-blocking — a deliberate author call to keep the long form is a valid choice; flagging because the duplication risk is real.

### What's Good

- **Whitelist over blacklist is the SEU-correct choice.** `fsm_state == L1 || L2 || L3 || POST_BRAKE` rejects every out-of-range value categorically, whereas the previous `fsm_state < FSM_BRAKE_L1` implicitly classified anything with the high bit flipped (e.g. `0x83` from an `FSM_OFF = 0x00` SEU) as "not braking" — fail-safe by coincidence, not by design. The rewrite moves this from coincidence to guarantee.
- **`isfinite()` at the boundary, not inside the PI math.** Catching the whole arithmetic chain in one place is the right lever — once `decel_target` and `a_ego` are known finite, every derived value (error, integral update, output) is finite too (modulo the bounded constants). Spot-checking inside the PI body would have been redundant and would have complicated the single-exit structure.
- **State reset is complete on the fail-safe branch.** `s_integral = 0.0F` and `s_prev_brake_pct = 0.0F` both cleared. If the guard trips and then clears (e.g. a transient NaN from a sensor glitch), the next cycle starts clean — no stale integrator, no jerk-limiter cross-contamination.
- **Unified guard preserves MISRA Rule 15.5.** `fsm_valid` and `inputs_finite` computed into named `const uint8_t` locals, then OR'd into the single-exit `if`. Readable, one return path, no early-exit special cases.
- **Honest coverage disclosure.** The 100% → 90.6% MC/DC drop is called out in the commit body with the right framing — the metric is responding correctly to new atomic conditions (9 added, 6 exercised by existing tests), and complementing the MC/DC suite is the V&V author's job, not the fix author's. Preserves ISO 26262-6 §5.4.8 independence.
- **Scope discipline.** One file (`src/execution/aeb_pid.c`), one function (`pid_brake_step`), one include (`<math.h>`). PI control law, jerk limiter, output clamping, alert module — all untouched. Pure hardening at the input boundary.
- **Pattern-matches the companion fixes.** Same `isfinite()` + single-exit + fail-safe-zero structure as `fix/uds-robustness` (#94) and `fix/can-encode-robustness` (#102). The cross-module V&V hardening is coherent, which makes the assessor's life easier.
- **`fault-pid-alert` goes 12/18 → 18/18.** Commit body reports the expected flip, matching the same delta that `fix/uds-robustness` produced for UDS. Independence-loop closure pending on the V&V author's re-run post-merge.

---

jessica-leoa:
## CI failure on `mcdc-pid-alert` — scope question for the team

The CI run after force-push failed the branch coverage gate:

```
PID+Alert coverage — stmt=100.00% branch=91.89% mcdc=91.89%
Error: Branch dropped to 91.89% (required ≥95%)
```

This is the exact scenario flagged in the commit body and PR description. Want to surface it explicitly and ask the team how to proceed — it's a process question, not a code question.

### Root cause (confirmed)

The unified guard introduces 9 new atomic conditions relative to the pre-patch structure:

- 4× whitelist equality checks (`fsm_state == FSM_BRAKE_L1/L2/L3/POST_BRAKE`)
- 2× `isfinite()` calls (`decel_target`, `a_ego`)
- 3× additional OR combinations in the unified `if`

Existing cases in `tests/test_pid_mcdc.c` were written against the pre-patch 2-term guard (`fsm_state < FSM_BRAKE_L1 || decel_target <= 0.0F`). They don't exercise independent toggling of each new condition, so the branch/MC/DC aggregate drops from 100% (23/23) to 91.89% (34/37 branches, 29/32 conditions).

Statement coverage remains at 100% (85/85) — no dead code introduced.

### Not a behavioral regression

- `make test` — all nominal suites pass, no regression
- `make fault-pid-alert` — PID 18/18 PASS (was 12/18, the 6 targeted asserts flipped), Alert 18/18 (unchanged)
- `make memory-pid-alert` — 0 errors (Valgrind + ASan + UBSan)

The uncovered conditions are exercised at the level of *behavioral outcomes* (fault injection confirms the guard triggers and resets state correctly), just not at the level of *structural MC/DC independence* that the gate measures.

### The process question

Per ISO 26262-6 §5.4.8 (principle of independence), V&V test authoring belongs to a role distinct from the fix author. The original task charter for this PR explicitly scoped testing infrastructure to the V&V author's next iteration, and @RianLinhares's V&V report §8 lists this as the expected cycle (detection → correction by author → re-validation by V&V author, including MC/DC complement).

Adding the complementary cases inside this PR would technically pass the gate, but it would:

1. Expand scope beyond what was charted and promised in the commit / PR description
2. Have the fix author write structural tests against their own patch (against the independence principle for ASIL-D)
3. Set a precedent that may conflict with the analogous situations coming from `fix/uds-robustness` (#94), `fix/can-encode-robustness` (#102), and any future robustness patches, all of which add conditions and would hit the same gate

### Options to decide

1. **Split the PR:** land this fix as-is, open a follow-up PR owned by @RianLinhares (or another V&V-role teammate) adding the MC/DC complement cases. Coverage restores in the follow-up. Preserves independence.

2. **Adjust the scope of this PR:** I add the cases here, documenting in the commit that scope was expanded in response to the gate. Fastest to merge, but with the caveats above.

3. **Temporarily waive the gate for this PR class:** team decides that robustness fixes adding new conditions get a per-PR waiver, tracked by an issue requiring MC/DC complement before the next release tag. Keeps velocity and independence.

4. **Other** — I'm deferring to the team on the right precedent here.

Happy to go with whichever route the team prefers. If option 1 or 3, no action needed from me on this branch (beyond this comment). If option 2, I'll start the test additions and push to this branch.

cc @RianLinhares @renatosfagundes

jessica-leoa:
Update addressing CI failure on mcdc-pid-alert gate:

- **a1e7322** `test(pid-mcdc)` — added 2 targeted MC/DC cases covering the 3 uncovered outcomes from the unified guard (C5-F, C6-F, C8-T via short-circuit). CI now green across all checks (lines 100%, branch ≥95%, conditions ≥95%).

Transparency: per commit body, adding MC/DC complement here is a pragmatic deviation from ISO 26262-6 §5.4.8 (V&V-author independence) under time constraints. @RianLinhares — please flag if a separate V&V-authored pass is still required for the audit trail; happy to revert this commit if so.

Ready for re-review when convenient. @renatosfagundes

erycaaf:
## Re-review — PR #100 · `fix/pid-robustness`

Force-push applied: dev sync (`f3e9624`) + MC/DC complement commit (`a1e7322` adds two targeted cases covering C5-F, C6-F, and C8-T via short-circuit — CI green, gate satisfied).

The test cases themselves read well: `test_mcdc_decel_target_nan` and `test_mcdc_a_ego_nan` each exercise exactly one new atomic condition, with explicit MC/DC reasoning in the header comment and real-world motivation tied to concrete upstream corruption scenarios (divide-by-zero, corrupted IMU). The code quality is not the concern.

The concern is **who wrote them**. 

### Must Fix

**1. `@RianLinhares` — please review `a1e7322` as if you had authored it.**

The independence principle of §5.4.8 exists because the fix author has a cognitive blind spot: they test the code they wrote the way they wrote it, not the way it could fail. A truly independent V&V pass would ask questions the fix author doesn't instinctively ask.

Please confirm in a reply on this PR either:

- **"Equivalent to what I would have written"** — the two cases cover the outcomes I would independently have targeted; I would not have added further cases for MC/DC at this level. → ready to approve.

- **"Adjustments needed"** — with concrete gaps. Specific prompts to consider:
  - compound case: both `decel_target` and `a_ego` non-finite simultaneously — does the short-circuit structure still give the right outcome? worth a third case?
  - ±Inf (not just NaN): `isfinite()` catches both, but the tests only feed NaN. Does the V&V report section require `+Inf` and `-Inf` exercised separately?
  - ordering assumption: the whitelist evaluates before the isfinite check in the current structure. If `fsm_state` is not in {L1/L2/L3/POST_BRAKE}, the guard trips regardless of finiteness. Is there a case that confirms the whitelist fires for an out-of-range FSM state paired with NaN inputs, so we know the ordering is intentional and not an artefact?
  - `pid_output_t` with stale prior state: does the integrator reset also get exercised when the NaN case trips mid-run, or only on init?

Your answer doesn't need to expand the test set if the current coverage satisfies the §5.4.8 audit bar in your judgment ...it just needs to be on record.

RianLinhares (approved):
## Independent V&V review of `a1e7322` — per @erycaaf's Must Fix

Going through `a1e7322` as if I hadn't seen this branch before, then
answering each of your four prompts directly.

### Verdict

**Equivalent to what I would have written — with caveats on process, not
code.** Approving.

Walked through both test bodies twice: once checking for MC/DC
independence (do they isolate C5 and C6 correctly?) and once trying to
find a gap I would have caught if I'd authored them first. On technical
content alone, the two cases match the V&V Report §8 specification
(`test_mcdc_decel_target_nan` → C5-F; `test_mcdc_a_ego_nan` → C6-F
with C5=T via finite `decel_target=4.0F` to force short-circuit past
the first operand). The MC/DC reasoning in the header comments is
stated correctly and would survive an auditor's question.

That's the code side. The process side deserves its own paragraph.

### On the four prompts

**1. Compound case (both `decel_target` and `a_ego` non-finite).**
Not needed for MC/DC. Once either atomic condition of an `&&` evaluates
false independently, §9.4.2 independence is satisfied. A "both NaN"
case would exercise the same outcome path as `test_mcdc_decel_target_nan`
(short-circuit on the first operand) and add no MC/DC evidence —
it could live in `test_pid_fault.c` as a robustness-under-compound-fault
assertion, but it's not a structural-coverage gap.

**2. ±Inf separation from NaN.** This is the one I would flag on a
formal audit. `isfinite()` treats NaN, +Inf, −Inf identically at the
C11 semantic level, but IEEE 754 arithmetic does not: `+Inf - offset`
can produce `+Inf`, `−Inf + 0.5F` can produce `−Inf`, and each has a
different UB signature if cast into `uint32_t` downstream. The guard
here rejects them all equivalently (correct), but the V&V report §7.1
does call out "NaN, +Inf, −Inf, and out-of-range finite" as four
distinct input classes. `test_pid_fault.c` already exercises the +Inf
and −Inf cases on `a_ego` (FAULT-A2, A3, B3), so the behavioural
coverage is there — but at the MC/DC level, only NaN is exercised.
**My judgment**: not a gate-blocker, since the structural contract
MC/DC measures is "atomic condition outcome toggles independently",
not "each physical IEEE 754 category triggers the outcome". If you
want belt-and-suspenders, a third case with `a_ego = INFINITY`
demonstrates the same C6-F outcome via a different mechanism and is
cheap to add. I would accept either.

**3. Ordering — whitelist before `isfinite` with out-of-range FSM +
NaN.** This is a good prompt and I checked the guard behaviour
explicitly. In the current structure, `fsm_valid == 0U` is the first
clause of the outer `if`, so a corrupted FSM state trips the guard
regardless of whether inputs are NaN — short-circuit evaluation never
reaches the `isfinite()` calls. The test coverage doesn't confirm
this ordering is intentional, so if the FSM whitelist ever gets
refactored (e.g. hoisted into a macro as @renatosfagundes suggested),
the ordering contract could silently break without any test catching
it. **My call**: `test_pid_fault.c` FAULT-C1/C2 already assert the
guard trips for `fsm_state=99` and `fsm_state=255` with finite inputs;
adding one assertion with `fsm_state=99 + decel_target=NaN + a_ego=NaN`
in `test_pid_fault.c` (not `test_pid_mcdc.c`) would pin the ordering.
I will take this on the follow-up V&V pass rather than expanding scope
here.

**4. Integrator reset mid-run (not just init).** Real gap. Both added
cases call `pid_init()` at the top, so they only prove the guard
produces zero on a clean state. They do not prove that the reset path
actually clears a *non-zero* `s_integral` if a NaN is injected after
the integrator has been accumulating. FAULT-D1 in `test_pid_fault.c`
covers "invalid FSM mid-run → reset", which is good, but not
"NaN mid-run → reset" — that's a distinct path through the same
reset code, on a different guard clause. **My call**: this one is
worth adding. Low cost, and closes a genuine blind spot.

### The process concern — on record

Code quality is not why I would have preferred Option 1. The risk
isn't that `a1e7322` is wrong; it's that it was written with the
guard implementation already in mind. The cases target exactly the
conditions the fix introduced, in exactly the order the code evaluates
them. That's efficient, but it means the tests answer the question
"do the branches I wrote execute?" rather than "do the branches I
wrote cover every failure mode that should fail safe?"

Prompt 4 above is an example of what that blind spot looks like in
practice — the reset path has two entry triggers (invalid FSM, NaN
input) and only one is exercised from a mid-run integrator state. I
caught it because I'm looking at the test from outside the fix
author's mental model; the fix author, having just implemented both
guards as one unified `if`, would reasonably assume "I reset in the
fail-safe branch, and the branch is exercised, therefore reset is
covered" — which is true structurally but incomplete behaviourally.

This is exactly the class of gap §5.4.8 independence is designed to
catch. It surfaced here anyway because the review-loop worked, but
the default-case — where nobody looks hard at test authorship — is
where the risk lives.

**Recommendation for this PR and future ones in this class:**

* Merge this PR with the current `a1e7322` (MC/DC gate satisfied,
  behavioural gaps manageable).
* I open a small follow-up PR on the V&V branch adding: (a) the
  `NaN + invalid FSM` ordering assertion to `test_pid_fault.c`, (b)
  the `NaN mid-run with pre-accumulated integrator` case to
  `test_pid_fault.c`, and optionally (c) ±Inf variants alongside the
  existing NaN cases in `test_pid_mcdc.c`. All authored by the V&V
  role, restoring the independence lane for audit.
* Going forward, when a fix PR would otherwise drop MC/DC below the
  95% gate, the default should be Option 1 from Jessica's four
  options — not Option 2 — so the time pressure doesn't land on the
  fix author. I'll raise this as a SOP note in the V&V board.

None of the above blocks this PR. Approving `a1e7322` on the narrow
MC/DC-gate question @erycaaf asked me to confirm. The V&V follow-up is
mine to own.

erycaaf (approved):
## Final review — PR #100

Approving. `@RianLinhares`'s independent pass confirms the MC/DC-gate contract is satisfied and validates `a1e7322` on the narrow question this review asked him to answer. 

### On the SOP recommendation

Strong endorsement of `@RianLinhares`'s third recommendation: going forward, when a fix PR would drop MC/DC below the 95% gate, the default should be Option 1 (split: fix PR merges with the gate lifecycle-adjusted, V&V author owns the complement PR) — not Option 2 (fix author expands scope). The review loop caught the independence-gap this time (prompt 4 surfaced a real hole), but as Rian notes, the default-case — where nobody cross-reviews — is where the risk lives. Worth capturing in the V&V board's SOP note so the precedent set here doesn't quietly propagate to future robustness PRs across the other modules.

---

PR #102, opened 2026-04-24
fix(can): guard encode_unsigned against NaN / +-Inf in float-to-uint32 cast

### LINKED ISSUE #101: [FIX] CAN: guard encode_unsigned against NaN / pm Inf in float-to-uint32 cast
## Context

`encode_unsigned` in `src/communication/aeb_can.c` converts a physical float value to the DBC raw form via:

```c
float32_t raw_f = (physical - offset) / factor;
uint32_t  raw   = 0U;

if (raw_f < 0.0F)  {
    raw = 0U;
} else {
    raw = (uint32_t)(raw_f + 0.5F);
}
```

## Bug

Surfaced by the independent ASIL-D V&V cross-validation conducted by @mphumapo2014 (see PR #89 consolidated report, Bug 1):

- IEEE 754 NaN compares false to everything, so the `raw_f < 0.0F` guard does not catch NaN. The cast `(uint32_t)(raw_f + 0.5F)` is then Undefined Behaviour per C11 section 6.3.1.4: "the behavior is undefined" when the float value cannot be represented in the integer type.
- plus/minus Inf hits the same path -- also UB on the cast.
- Very large finite floats (greater than UINT32_MAX) also produce UB on the cast.
- Division by zero in `(physical - offset) / factor` produces Inf, same UB fate.

UBSan confirmed the violation on the fault-injection suite in `reports/vv_can/memory_safety/` during PR #89's validation.

## Consequence

Under an upstream arithmetic error (SEU in `pid_out->brake_bar` RAM, `clamp_f32` passing NaN through because NaN comparisons are all false, sensor saturation producing Inf), the CAN TX path emits an undefined frame. On GCC 14 / x86-64 the UB happens to settle on a fixed pattern, but that is toolchain-dependent and not a safety guarantee.

### `src/communication/aeb_can.c`

- Add `#include <math.h>` for `isfinite()`.
- Rewrite `encode_unsigned` to:
  - Reject non-finite `physical` with deterministic zero.
  - Guard against non-finite `raw_f` (catches division-by-zero when `factor = 0`).
  - Saturate at `UINT32_MAX` for values that exceed the cast range.
  - Preserve the existing negative-clamp behaviour (raw = 0).
  - Single exit point (MISRA C:2012 Rule 15.5).

### Behavioural contract

- Legitimate inputs in DBC range produce identical output (nominal suite stays green).
- NaN / plus/minus Inf / out-of-range float -> deterministic 0 or UINT32_MAX saturation, never UB.

## Acceptance

- `make test` -- all nominal suites pass (no regression).
- `make misra` / `make misra-uds` / `make misra-decision` / `make misra-pid-alert` -- no new MISRA violations.
- `make fault-pid-alert` under UBSan -- the UB line from `encode_unsigned` no longer appears in the suite that exercises encode with corrupted `brake_bar`.
- Local reproduction: compile `aeb_can.c` with `-fsanitize=undefined` and feed a NaN through `encode_unsigned` directly -- no runtime error reported.

## Not in scope (separate PRs)

- Adding defence-in-depth validation in TX callers (`can_tx_brake_cmd`, `can_tx_fsm_state`) -- the fix at the `encode_unsigned` boundary is the correct layer.
- Boolean normalisation on `can_get_rx_data` output -- tracked separately if needed.

## Why this lands now

The bug is documented, UBSan-confirmed, and the fix preserves all existing behaviour for legitimate inputs. Lourenco's V&V (PR #89) explicitly deferred the patch to `fix/can-encode-robustness` authored by the module owner per ISO 26262-6 section 5.4.8 independence principle.

renatosfagundes:
# Summary
- Guard `encode_unsigned` in `src/communication/aeb_can.c` against NaN / plus-minus-Inf / out-of-range float inputs that caused Undefined Behaviour on the `(uint32_t)(raw_f + 0.5F)` cast (C11 section 6.3.1.4).
- Replace the fail-safe-by-coincidence pattern (`raw_f < 0.0F` does not catch NaN because NaN comparisons are all false) with explicit `isfinite()` checks on input and post-arithmetic, plus upper-bound saturation at UINT32_MAX.
- No behavioural change for legitimate inputs: all 195/195 nominal tests stay green.

# Related Issue
Closes #101

Surfaces the fix deferred by PR #89 consolidated report (Bug 1). Same defensive pattern adopted by #94 (UDS float-to-uint16 validation) and #100 (PID `isfinite` guards).

# Change Type
- [ ] feat
- [x] fix
- [ ] docs
- [ ] test
- [ ] ci
- [ ] refactor/chore

# AEB Areas Affected
- [ ] Perception
- [ ] Decision Logic
- [ ] Driver Alerts
- [ ] Braking Control
- [ ] State Machine
- [x] CAN Interface
- [ ] UDS Diagnostics
- [ ] Integration
- [ ] Documentation only

# Requirements Impacted
## Functional Requirements
- FR-CAN-003 (DBC signal encoding) -- now robust to invalid float inputs.

- NFR-SAF-ROB (robustness against invalid inputs).
- NFR-COD-001 (MISRA C:2012 -- Rule 15.5 single exit point preserved).

# Artifacts Updated
- [ ] Requirements document
- [ ] Modeling document
- [x] C source code
- [ ] Tests
- [ ] CI workflow
- [ ] CHANGELOG

# Validation Evidence
- [x] Local build passes (`make build` -- zero warnings, `-Wall -Wextra -Wpedantic -std=c99 -O2`)
- [x] Automated tests pass (`make test` -- 195/195 across all 8 suites, no regression)
- [ ] CI passes (automatic once opened)
- [x] Traceability updated (`Closes #101`, referenced in commit message)
- [x] Safety-relevant impact assessed (see Evidence)
- [ ] Documentation updated

### The bug (as flagged by the V&V cross-validation)

```c
/* before — raw_f < 0.0F lets NaN fall through, cast is UB */
float32_t raw_f = (physical - offset) / factor;
if (raw_f < 0.0F) {
    raw = 0U;
} else {
    raw = (uint32_t)(raw_f + 0.5F);   /* UB on NaN, +-Inf, huge finite */
}
```

Four UB-inducing input classes:
1. NaN (upstream `clamp_f32` silently passes NaN through because NaN comparisons are false).
2. plus-minus Inf (sensor saturation, arithmetic overflow).
3. Finite floats greater than UINT32_MAX.
4. `factor = 0` producing Inf via division.

### The fix

```c
uint32_t raw = 0U;

if (isfinite(physical)) {
    const float32_t raw_f = (physical - offset) / factor;

    if (isfinite(raw_f) && (raw_f >= 0.0F)) {
        if (raw_f > 4294967040.0F) {
            raw = UINT32_MAX;              /* saturate */
        } else {
            raw = (uint32_t)(raw_f + 0.5F);
        }
    }
}

return raw;   /* single exit */
```

Output contract:
- Finite in-range input -> identical to the pre-fix value (nominal suite unchanged).
- NaN / plus-minus Inf -> 0 (deterministic).
- Negative finite -> 0 (legacy behaviour preserved).
- Finite greater than UINT32_MAX ceiling -> UINT32_MAX (deterministic saturation).
- `factor = 0` -> isfinite(raw_f) catches the Inf -> 0.

### Test results

```
test_smoke:        14/14
test_can:          52/52
test_perception:   25/25
test_decision:      9/9
test_pid:          19/19
test_alert:        27/27
test_uds:          23/23
test_integration:  26/26
---
total:            195/195   (no regression)
```

### Local UBSan on Windows/MinGW

The MinGW 15 toolchain on the author's Windows host does not ship `libubsan`, so UBSan verification was not runnable locally. The fix logic is straightforward (finite-guard + saturate before cast) and the CI Ubuntu 24.04 runner exercises `-fsanitize=undefined` via `make memory-uds` / `make memory-pid-alert` / `make memory-decision`. Expect the `encode_unsigned` line to stop appearing in the fault-suite UBSan log once PR #100 (PID robustness, which exercises the corrupted `brake_bar` path) and this PR both land.

# Reviewer Notes

- Defence-in-depth only at `encode_unsigned` -- no changes to `can_tx_brake_cmd` / `can_tx_fsm_state` callers. Fixing at the boundary is the correct layer: every float-input DBC encode path goes through this one function.
- `UINT32_MAX` comes from `<stdint.h>` via `aeb_types.h` -- no new header dependency.
- `4294967040.0F` is `UINT32_MAX - 255`, the largest float32 that still rounds cleanly into uint32 (one-line comment in the source explains this).
- Pattern is deliberately consistent with #94 (UDS) and #100 (PID) so all three modules use the same style of float-input hardening.

# Risks / Open Points

- UBSan log change won't be visible until PR #100 (PID robustness) merges and `fault-pid-alert` exercises the corrupted `brake_bar` path through the encoder. This PR is the receiver-side fix; PR #100 produced the sender-side fix.
- `can_tx_fsm_state` calls `encode_unsigned(ttc_thresh, 0.1F, 0.0F)` with `ttc_thresh` derived from a `switch` with hardcoded floats (`TTC_WARNING`, `TTC_BRAKE_L1`, ...) -- those paths were already safe by construction. The fix is here for defence-in-depth.
- MISRA: `isfinite()` is standard C99 but its expansion can produce a comparison that cppcheck sometimes flags as 10.1 (implicit type conversion). If the gate trips, an `(x != 0)` wrap on the `isfinite()` call is the standard remediation -- will address in the review cycle if needed.

### DIFF
--- a/src/communication/aeb_can.c
+++ b/src/communication/aeb_can.c
@@ -78,27 +78,10 @@ static uint32_t encode_unsigned(float32_t physical,
                                 float32_t factor,
                                 float32_t offset)
 {
-    /* Robust float-to-uint32 conversion for DBC signal packing.
-     *
-     * IEEE 754 NaN compares false to everything, so a naive
-     * `raw_f < 0.0F` guard lets NaN fall through to
-     * `(uint32_t)(raw_f + 0.5F)` — undefined behaviour per C11
-     * §6.3.1.4. ±Inf and very large finite floats hit the same UB.
-     * Division by zero in (physical - offset) / factor also produces
-     * ±Inf → UB.
-     *
-     * This function guarantees deterministic output for any float
-     * input: finite in-range values encode normally; non-finite or
-     * out-of-range values saturate to 0 (negative / NaN) or
-     * UINT32_MAX (positive overflow) so the resulting CAN frame is
-     * always well-defined — never UB, never toolchain-dependent.
-     *
-     * Upper bound 4294967040.0F (= UINT32_MAX - 255) is the largest
-     * float32 value that, after +0.5F rounding, still casts into
-     * uint32 without overflow. Values above that saturate.
-     *
-     * Single exit point (MISRA C:2012 Rule 15.5).
-     * Surfaced by V&V cross-validation on PR #89 (Bug 1). */
+    /* NaN / ±Inf / overflow on the float→uint32 cast is UB per C11
+     * §6.3.1.4; saturate at 4294967040.0F (UINT32_MAX − 255, largest
+     * float32 that rounds cleanly into range). See commit body for
+     * full rationale. */
     uint32_t raw = 0U;
 
     if (isfinite(physical))
--- a/src/communication/aeb_can.c
+++ b/src/communication/aeb_can.c
@@ -14,6 +14,7 @@
 
 #include "aeb_can.h"
 #include "can_hal.h"   /* Zephyr CAN HAL stub / real driver */
+#include <math.h>      /* isfinite() — NaN / ±Inf guard on encode_unsigned */
 #include <string.h>
 
 /* ═══════════════════════════════════════════════════════════════════════
@@ -77,16 +78,44 @@ static uint32_t encode_unsigned(float32_t physical,
                                 float32_t factor,
                                 float32_t offset)
 {
-    float32_t raw_f = (physical - offset) / factor;
-    uint32_t  raw   = 0U;
+    /* Robust float-to-uint32 conversion for DBC signal packing.
+     *
+     * IEEE 754 NaN compares false to everything, so a naive
+     * `raw_f < 0.0F` guard lets NaN fall through to
+     * `(uint32_t)(raw_f + 0.5F)` — undefined behaviour per C11
+     * §6.3.1.4. ±Inf and very large finite floats hit the same UB.
+     * Division by zero in (physical - offset) / factor also produces
+     * ±Inf → UB.
+     *
+     * This function guarantees deterministic output for any float
+     * input: finite in-range values encode normally; non-finite or
+     * out-of-range values saturate to 0 (negative / NaN) or
+     * UINT32_MAX (positive overflow) so the resulting CAN frame is
+     * always well-defined — never UB, never toolchain-dependent.
+     *
+     * Upper bound 4294967040.0F (= UINT32_MAX - 255) is the largest
+     * float32 value that, after +0.5F rounding, still casts into
+     * uint32 without overflow. Values above that saturate.
+     *
+     * Single exit point (MISRA C:2012 Rule 15.5).
+     * Surfaced by V&V cross-validation on PR #89 (Bug 1). */
+    uint32_t raw = 0U;
 
-    if (raw_f < 0.0F)
+    if (isfinite(physical))
     {
-        raw = 0U;
-    }
-    else
-    {
-        raw = (uint32_t)(raw_f + 0.5F);  /* round to nearest */
+        const float32_t raw_f = (physical - offset) / factor;
+
+        if (isfinite(raw_f) && (raw_f >= 0.0F))
+        {
+            if (raw_f > 4294967040.0F)
+            {
+                raw = UINT32_MAX;           /* saturate at cast ceiling */
+            }
+            else
+            {
+                raw = (uint32_t)(raw_f + 0.5F);  /* round to nearest */
+            }
+        }
     }
 
     return raw;
--- a/tests/test_can.c
+++ b/tests/test_can.c
@@ -313,6 +313,96 @@ TEST(test_tx_fsm_period)
 }
 
 /* ═══════════════════════════════════════════════════════════════════════
+ *  TESTS: encode_unsigned robustness — exercised through can_tx_brake_cmd,
+ *  which feeds pid_out->brake_bar into encode_unsigned for the 15-bit
+ *  BrakePressure field at bits 1..15. Lock the post-fix contract so any
+ *  future regression at the cast boundary is caught at unit level instead
+ *  of waiting for fault-injection.
+ *
+ *  The legitimate-input path is already covered by test_tx_brake_cmd
+ *  above; these four tests cover the four UB-inducing input classes the
+ *  cross-validation report flagged on PR #89.
+ *
+ *  NaN / ±Inf -> encode_unsigned returns 0 -> BrakePressure raw == 0.
+ *  Finite > UINT32_MAX ceiling -> saturates to UINT32_MAX -> 15-bit
+ *  field masked to 0x7FFF (all-ones).
+ * ═══════════════════════════════════════════════════════════════════════ */
+
+/** Helper — extract the 15-bit BrakePressure field from a captured frame. */
+static uint32_t extract_brake_pressure(const tx_record_t *tx)
+{
+    return can_unpack_signal(tx->data, 1U, 15U);
+}
+
+TEST(test_tx_brake_cmd_nan_brake_bar)
+{
+    can_state_t state;
+    can_hal_test_reset();
+    (void)can_init(&state);
+
+    pid_output_t pid = { .brake_pct = 0.0F, .brake_bar = NAN };
+    fsm_output_t fsm = { .fsm_state = (uint8_t)FSM_BRAKE_L1 };
+
+    int32_t rc = can_tx_brake_cmd(&state, &pid, &fsm);
+    ASSERT_EQ(rc, CAN_OK);
+    ASSERT_EQ(can_hal_test_get_tx_count(), 1U);
+
+    const tx_record_t *tx = can_hal_test_get_tx(0U);
+    ASSERT_EQ(extract_brake_pressure(tx), 0U);
+}
+
+TEST(test_tx_brake_cmd_pos_inf_brake_bar)
+{
+    can_state_t state;
+    can_hal_test_reset();
+    (void)can_init(&state);
+
+    pid_output_t pid = { .brake_pct = 0.0F, .brake_bar = INFINITY };
+    fsm_output_t fsm = { .fsm_state = (uint8_t)FSM_BRAKE_L1 };
+
+    int32_t rc = can_tx_brake_cmd(&state, &pid, &fsm);
+    ASSERT_EQ(rc, CAN_OK);
+
+    const tx_record_t *tx = can_hal_test_get_tx(0U);
+    ASSERT_EQ(extract_brake_pressure(tx), 0U);
+}
+
+TEST(test_tx_brake_cmd_neg_inf_brake_bar)
+{
+    can_state_t state;
+    can_hal_test_reset();
+    (void)can_init(&state);
+
+    pid_output_t pid = { .brake_pct = 0.0F, .brake_bar = -INFINITY };
+    fsm_output_t fsm = { .fsm_state = (uint8_t)FSM_BRAKE_L1 };
+
+    int32_t rc = can_tx_brake_cmd(&state, &pid, &fsm);
+    ASSERT_EQ(rc, CAN_OK);
+
+    const tx_record_t *tx = can_hal_test_get_tx(0U);
+    ASSERT_EQ(extract_brake_pressure(tx), 0U);
+}
+
+TEST(test_tx_brake_cmd_overflow_brake_bar)
+{
+    can_state_t state;
+    can_hal_test_reset();
+    (void)can_init(&state);
+
+    /* 1e20 / 0.1 = 1e21 — far above the 4294967040.0F cast ceiling.
+     * encode_unsigned must saturate at UINT32_MAX; can_pack_signal
+     * then masks to the 15-bit field width = 0x7FFF (all-ones). */
+    pid_output_t pid = { .brake_pct = 0.0F, .brake_bar = 1.0e20F };
+    fsm_output_t fsm = { .fsm_state = (uint8_t)FSM_BRAKE_L1 };
+
+    int32_t rc = can_tx_brake_cmd(&state, &pid, &fsm);
+    ASSERT_EQ(rc, CAN_OK);
+
+    const tx_record_t *tx = can_hal_test_get_tx(0U);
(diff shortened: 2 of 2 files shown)


RianLinhares (dismissed):
LGTM ✅ with a few non-blocking notes for the follow-up cycle.

The UB analysis is correct and the fix addresses the four failure
classes the consolidated V&V report on #89  flagged:

1. **NaN** — `isfinite(physical)` at the entry catches it; the old
   `raw_f < 0.0F` guard let it fall through because IEEE 754 NaN
   compares false to every relation.
2. **±Inf on input** — same entry guard.
3. **Inf produced by arithmetic** (`factor = 0`, or `(physical -
   offset) / factor` overflowing) — the second `isfinite(raw_f)`
   catches this; without it, `physical = finite` + `factor = 0`
   would still hit the UB cast.
4. **Finite > UINT32_MAX** — the `4294967040.0F` ceiling is correct.
   I double-checked: `UINT32_MAX = 4294967295` cannot be represented
   exactly in float32 (24-bit mantissa), and `4294967040 = UINT32_MAX
   - 255` is the largest float32 that, after `+0.5F` rounding, still
   fits into uint32 without overflow. Good catch.

Pulled the branch and confirmed `make build` (`-Wall -Wextra
-Wpedantic -std=c99`) is clean and `make test` reports 195/195. No
regression on the nominal suite.

The pattern is consistent with what we did on `aeb_uds.c` (#94) and
what #100 landed on `aeb_pid.c`: `isfinite()` + saturate + single
exit. Glad to see all three modules converging on the same style —
makes it easier for future contributors to recognise "this is how
we harden float boundaries" in this codebase.

**Non-blocking notes for the follow-up cycle** (not asking to hold
the PR for any of these):

* **Positive unit tests for the new guards.** The nominal suite
  didn't exercise NaN / ±Inf / `factor=0` / finite > UINT32_MAX on
  `encode_unsigned` before — if it had, we'd have caught the bug at
  unit level instead of waiting for fault injection. Might be worth
  adding four asserts to `tests/test_can.c` (or a new
  `test_can_fault.c` following the per-module pattern) so the
  corrected behaviour becomes a fixed, regression-detectable
  contract and doesn't rely on #100's fault path exercising it
  indirectly.

* **MC/DC coverage on `aeb_can.c`.** The function grew three new
  branches (`!isfinite(physical)`, `!isfinite(raw_f) || raw_f 
  0.0F`, `raw_f > 4294967040.0F`). The nominal suite only exercises
  the all-true path, so structural coverage on this file should dip
  until a `*_mcdc` or `*_fault` suite lands. If `vv-can` is planned
  the same way as `vv-uds` / `vv-decision`, worth making sure the
  complementary cases land together with the module's V&V workflow.

* **MISRA gate.** You flagged 10.1 and potentially 21.11 on
  `isfinite()`. #94 already shipped `<math.h>` + `isfinite()` on
  UDS through the cppcheck gate without issues, so the precedent is
  there — should be fine, but worth confirming `make misra-*` still
  passes on the touched file when the CAN module gets its own V&V
  workflow.

* **Semantic visibility of substituted values.** The fix converts
  NaN → 0 and overflow → UINT32_MAX silently at the CAN-TX boundary,
  which is correct for fail-safe determinism but makes an upstream
  NaN indistinguishable from a legitimate `physical = 0` on the bus.
  The UDS DTC layer (C1001 / C1004 / C1006) is the right place to
  catch the upstream fault before it reaches here, so I think this
  is fine — just worth a sentence in the consolidated V&V report
  noting the defence-in-depth layering (sensor → DTC → CAN encode)
  and the tradeoff.

Approving. None of the above blocks merge.

erycaaf:
## Review — PR #102 · `fix/can-encode-robustness`

### Should Fix

**1. Inline doc-block is ~14 lines and duplicates the commit body.**

The new comment in `encode_unsigned` covers IEEE 754 NaN behaviour, C11 UB on cast, the saturation ceiling derivation, the single-exit MISRA reference, and the PR #89 cross-validation link. All of that is already in the commit message at equal or greater depth. Two places for the same rationale is a maintenance trap — drift is inevitable, and the V&V report text will add a third copy.

Per the project's lean-comments convention, the inline should collapse to the non-obvious bits only:

```c
/* NaN / ±Inf / overflow in the float→uint32 cast is UB per C11 §6.3.1.4;
 * saturate at 4294967040.0F (UINT32_MAX − 255, largest float32 that
 * rounds cleanly into range). See commit body for full rationale. */
```

The "(see commit body)" pointer gives the future reader one git-blame hop instead of two. The C11 reference and the 4294967040.0F derivation are the bits that are genuinely non-obvious at the line; everything else — MISRA 15.5, fail-safe philosophy, PR #89 traceability — belongs in the commit/report layer, not here.

Non-blocking — stylistic alignment with the rest of `aeb_can.c` (which uses concise doxygen headers and sparse inline comments), not a correctness issue.

**2. Branch is two commits behind `development`.**

`git merge-base origin/development origin/fix/can-encode-robustness` → `e7589a7` (the `feat(can-uds)` integration commit). Current dev is at `6aeb0b4`. The missing two commits are from PR #103 (the `fault-uds` promote-to-blocking flip) and touch `.github/workflows/vv-uds.yml` + `Makefile`. No file overlap with this PR's `src/communication/aeb_can.c` edit, so merge-time conflict is not expected — but a `git merge origin/development` before the merge keeps the auditor's history clean and confirms `make fault-uds` on `development` still goes green with the new blocking gate.

### What's Good

- **Fix is layered correctly.** `isfinite(physical)` catches the input before the arithmetic; `isfinite(raw_f)` catches the intermediate when `(physical − offset) / factor` amplifies to ±Inf (e.g. `factor == 0.0F` division); the `raw_f > 4294967040.0F` check saturates before the cast can trip UB. All three guards are needed — dropping any one leaves a hole.
- **Saturation ceiling is mathematically precise.** `UINT32_MAX = 2³² − 1 = 4294967295`; float32 precision at that magnitude is `2⁸ = 256` (23-bit mantissa + implicit 1). The largest representable float32 strictly below `UINT32_MAX` is `4294967040 = 2³² − 256`. Any float32 above that rounds to `2³²` which is out of `uint32_t` range. The chosen constant is exactly the tight bound, not a conservative approximation.
- **Fail-safe semantics compose correctly with DBC packing.** When `raw = UINT32_MAX` meets `can_pack_signal(data, start_bit, length, raw)`, the bitmask reduces it to all-ones within the signal's width (`0xFFFF` for a 16-bit signal). The receiver sees "maximum physical value reported" rather than implementation-defined garbage — the right degradation for a sensor-saturation or SEU scenario.
- **Single exit preserved (MISRA Rule 15.5).** The nested `if` structure keeps `return raw;` as the sole exit. The variable is explicitly initialised to `0U` up top so every un-entered branch produces a deterministic zero — no indeterminate state leaks out.
- **No behaviour change for legitimate inputs.** The positive-finite-in-range path remains exactly `(uint32_t)(raw_f + 0.5F)` — same rounding, same result. The 195-nominal-tests-green claim in the commit body is consistent with the diff; the fix is additive in the defensive-input space, not modifying the happy-path arithmetic.
- **Companion-fix pattern acknowledged.** Commit body notes the parallel fixes in PR #94 (UDS uint16 cast) and PR #100 (PID `decel_target` / `a_ego` finiteness). That cross-module reference helps an assessor see this as a coordinated project-wide hardening, not a one-off.
- **Scope discipline.** One static function, one include, no header change, no test change. Pure module-level patch. The cross-validator's V&V stack is independent — `make fault-can` on this branch should flip from `30/31 PASS + 1 UBSan` to `31/31 PASS + 0 UBSan`, which the V&V author re-runs post-merge.
- **MISRA surface unchanged.** The rewrite doesn't introduce any new rules to deviate: `isfinite` is a Directive 4.9 macro (allowed), no new pointer arithmetic, no shadowing, single exit preserved, comparison to non-zero remains explicit.

---

renatosfagundes:
Thanks for the reviews, @erycaaf and @RianLinhares. All actionable items addressed across three commits on top of the original fix:

| # | Reviewer | Item | Fix | Commit |
|---|---|---|---|---|
| 1 | @erycaaf | Branch 2 commits behind `development` | Merged `origin/development` (no conflicts; PR #103's changes touch `vv-uds.yml` + `Makefile`, no overlap with `aeb_can.c`). | `97fcfdc` |
| 2 | @erycaaf | Inline doc-block (~14 lines) duplicates the commit body | Collapsed to 4 lines covering the genuinely non-obvious bits (C11 §6.3.1.4 reference + `4294967040.0F` derivation) with a "see commit body" pointer. | `7fd78e7` |
| 3 | @RianLinhares | Add positive unit tests for the four UB-inducing input classes | Added `test_tx_brake_cmd_{nan,pos_inf,neg_inf,overflow}_brake_bar` to `tests/test_can.c`. `encode_unsigned` is `static`, so the tests exercise it through `can_tx_brake_cmd` (the only public path that takes a float input → encode_unsigned, via the 15-bit BrakePressure field) and decode the captured TX bytes with `can_unpack_signal`. | `a725647` |

### Test count

```
test_smoke:        14/14
test_can:          61/61   (+9 from the 4 new tests)
test_perception:   25/25
test_decision:      9/9
test_pid:          19/19
test_alert:        27/27
test_uds:          23/23
test_integration:  26/26
---
total:            204/204  (was 195/195)
```

Zero warnings under `-Wall -Wextra -Wpedantic -std=c99 -O2`.

### Items consciously left for the V&V cycle (not on this PR)

- **MC/DC complementary cases on `aeb_can.c`** — the function grew three new branches; nominal-only coverage will dip until the module's `*_mcdc` / `*_fault` complementary suite exercises them. The four positive tests added here lock the all-true path *and* the four defensive paths, but a structural complementary suite is a different artefact and belongs with the module's V&V workflow.
- **MISRA gate confirmation on `<math.h>` + `isfinite()`** — relevant when the CAN module's V&V workflow lands; PR #94 already shipped the same pattern through the UDS MISRA gate without issue, so the precedent is in place.
- **Defence-in-depth note in the V&V consolidated report** — the silent NaN -> 0 substitution at the CAN-TX boundary is correct fail-safe behaviour, but worth a sentence in the report explaining the layering (sensor -> DTC -> CAN encode) so an assessor sees it as deliberate, not a missing diagnostic. Documentation task for the V&V author.

Ready for another look whenever convenient.

erycaaf (approved):
lgtm

RianLinhares (approved):
LGTM

PR #62, opened 2026-04-11
fix: address PR #61 review — gitignore, task nomenclature, CI triggers

erycaaf:
## Summary
- Add 6 missing test binaries (`test_perception`, `test_decision`, `test_pid`, `test_alert`, `test_uds`, `test_integration`) to `.gitignore`
- Replace 7 occurrences of internal "Task A/B/C" nomenclature with actual module names (Perception, FSM, PID, Alert) in `aeb_can.h`, `aeb_ttc.c`, and `test_perception.c`
- Add `main` branch to CI workflow `push` and `pull_request` triggers so releases are validated

## Related Issue
Ref #61

## Change Type
- [x] fix
- [x] ci

## AEB Areas Affected
- [x] Perception
- [x] Decision Logic
- [x] CAN Interface
- [x] Integration

### Functional Requirements
- N/A (no functional changes — comments and CI config only)

### Non-Functional Requirements
- NFR-COD-005..007 (code quality: removal of internal nomenclature from documentation)
- NFR-VAL-001 (CI coverage: main branch now validated)

## Artifacts Updated
- [x] C source code
- [x] Tests
- [x] CI workflow

## Validation Evidence
- [x] Local build passes
- [x] Automated tests pass
- [ ] CI passes
- [x] Safety-relevant impact assessed
- [x] Documentation updated

### Evidence
- Changes are limited to comments, `.gitignore`, and CI config — no functional code modified
- `make build` — zero-warning gate unaffected
- `make test` — all 209 assertions pass (no logic changes)
- Files modified: `.gitignore`, `include/aeb_can.h`, `src/decision/aeb_ttc.c`, `tests/test_perception.c`, `.github/workflows/ci.yml`
- No impact on CCRs / CCRm / CCRb scenarios

## Reviewer Notes
- All changes are non-functional: comments, `.gitignore` entries, and CI trigger config
- Verify that the  references listed in the PR #61 review are fully addressed

## Risks / Open Points
- None — no functional code changes

PR #116, opened 2026-04-27
feat(sil): UDS Diagnostic scenario — client, REST endpoints, dashboard panel

renatosfagundes:
# Summary

- Adds the **UDS Diagnostic scenario** to the SIL: Python TCP-CAN client (`uds_client.py`), five REST endpoints in the FastAPI app (`/uds/health`, `/uds/read_did/{did}`, `/uds/snapshot`, `/uds/clear_dtc`, `/uds/routine_control`), a slide-in dashboard panel that polls live DIDs at 5 Hz and exposes Clear-DTC + AEB enable/disable, plus the `uds_diagnostic` scenario entry.
- Adds a **`ccrs_50_no_aeb` control scenario** that fires UDS RoutineControl(0x0301, 0) right after the ECU restart, so the FSM stays in OFF for the entire run — useful as a comparison-with-AEB baseline.
- Adds an **FSM-state debounce** in the dashboard frontend (cascade pills + UDS panel) that holds the last committed non-OFF state for 500 ms before surfacing OFF. Filters transient flickers (perception ROC during velocity ramp, ECU restart settle) without hiding persistent OFF (AEB disabled, real sensor fault).
- 15 unit tests for `uds_client.py` driven against a local fake TCP relay, covering wire-level encoding/decoding, negative responses, timeouts, DID-echo mismatches, and concurrent request serialisation. All green locally.

# Related Issue

Builds on #97 (UDS over CAN in production C — already merged into `development`). Branched off `feat/sil-gazebo-docker` (PR #120).

# Change Type

- [x] feat
- [ ] fix
- [ ] docs
- [ ] test
- [ ] ci
- [ ] refactor/chore

# AEB Areas Affected

- [ ] Perception
- [ ] Decision Logic
- [ ] Driver Alerts
- [ ] Braking Control
- [ ] State Machine
- [ ] CAN Interface
- [x] UDS Diagnostics
- [x] Integration

# Requirements Impacted

## Functional Requirements

- **FR-UDS-001 / 003 / 004 / 005** — surfaces the production C UDS server end-to-end via HTTP. Request and response stay on the production side; this PR adds the *consumer* path: dashboard → FastAPI → TCP-CAN relay → ECU on `0x7DF`/`0x7E8`.

- **NFR-VAL-005** — adds a manually-triggered SIL scenario for diagnostic verification, complementing the Euro NCAP CCRs/CCRm/CCRb scenarios.

# Artifacts Updated

- [ ] Requirements document
- [ ] Modeling document
- [x] C source code
- [x] Tests
- [ ] CI workflow
- [x] README / SCM documentation
- [ ] CHANGELOG

(C source code box ticked because the new SIL UDS path exercises FR-UDS-005 end-to-end against unmodified production C.)

# Validation Evidence

- [x] Local build passes
- [x] Automated tests pass
- [x] Live SIL stack walkthrough completed
- [ ] CI passes *(no SIL-specific workflow exists yet)*
- [x] Traceability preserved
- [x] Documentation updated

## Evidence

**Unit tests**

```
$ cd sil/docker/api && python -m pytest tests/ -v
============================ 15 passed in 0.48s =============================
```

**Live SIL stack walkthrough**

`docker compose up --build` brought up canbus + gazebo + aeb-ecu + api + dashboard. Verified:

- `/uds/health` → connected
- `uds_diagnostic` scenario → ego stationary, dashboard's UDS panel polls TTC=10.0s, FSM_STATE=STANDBY, brake=0% live at 5 Hz
- `ccrs_50_no_aeb` scenario → ECU restart + UDS RoutineControl(0x0301, 0) → FSM stays in OFF the whole run, ego closes on target with no AEB intervention (control case)
- `ccrs_50` scenario (AEB enabled) → FSM cascade STANDBY → WARNING → BRAKE_L1/L2/L3 → POST_BRAKE observed live in both the cascade pills AND the UDS panel
- Clear DTCs (0x14) and AEB enable/disable toggle exercised — log row colour-coded green for success
- Negative-response paths exercised (e.g., requesting an unsupported DID) → red log row, panel stays alive

**Frame-format compatibility verified**: `uds_client.py` 16-byte `<IBxxx><8B>` matches `sil/docker/canbus/can_tcp_server.py:FRAME_SIZE = 16`.

**Affected scenarios**: none of CCRs/CCRm/CCRb altered. Two new scenarios:
- `uds_diagnostic` — quiescent baseline with ego stationary; demo flow for UDS panel.
- `ccrs_50_no_aeb` — identical to ccrs_50 but `disable_aeb: true` flag triggers UDS-driven AEB disable on scenario start. The control case for "what would happen without AEB" comparisons.

# Reviewer Notes

Three things worth focusing on:

1. **Frame-format compatibility with `canbus`.** `uds_client.py` uses the same 16-byte `<IBxxx><8B>` layout as `sil/docker/canbus/can_tcp_server.py` and `sil/aeb_gazebo/src/can_tcp_client.py`. Confirm the layout in `_send_request` (line ~120 of `uds_client.py`) matches the relay's `_pack_frame`.

2. **Lazy UDS client construction in `scenario_api.py`.** `_get_uds_client()` defers the TCP connection until the first `/uds/*` request to avoid blocking app startup when the canbus container is still coming up. Connection errors return HTTP 503 and the next request will retry.

3. **`disable_aeb` flag in scenarios.yaml + `/scenarios/start` UDS hook.** The flag fires UDS RoutineControl(0x0301, 0) **after** `restart_aeb_ecu()` returns, because UDS state on the Zephyr ECU is RAM-only and a fresh container always boots with `aeb_enabled = 1` (per `aeb_core_init` in `aeb_core.c`). Other scenarios are unaffected — only those with the YAML opt-in get the disable.

4. **FSM-state debounce in `index.html`** (the `fsmDebounce` helper). Filters sub-500 ms OFF flickers from perception ROC during velocity ramp and ECU restart settle, *without* hiding persistent OFF. The "raw = N" subtitle on the UDS panel's FSM card always shows the un-debounced ECU value for debugging.

# Risks / Open Points

1. **No SIL CI workflow exists** — the unit tests for `uds_client.py` will be picked up if/when a SIL workflow is added; today they are run-locally only.

2. **Dashboard FSM debounce is presentation-layer only** — the underlying `/live` and `/uds/snapshot` payloads still carry the raw ECU state. Anything reading those endpoints programmatically (test scripts, external tools) sees the un-debounced values. Document this in the API contract if/when one is published.

3. **`disable_aeb` flag is currently honoured on `/scenarios/start` only** — `/scenarios/restart` does not re-fire the disable, so a Restart click on `ccrs_50_no_aeb` would resurrect AEB. Easy follow-up; not on the critical path.

4. **`ccrs_50_no_aeb` always collides at 50 km/h** — that's the expected control behaviour, but the dashboard's collision banner ("STOPPED · gap N m") doesn't fire for collision events; only for stops. UI nicety, not a functional issue.

5. **CCRb_d6_g40 unchanged** — still collides at 19.7 km/h with AEB enabled (carried over from PR #120). FSM-tuning conversation, separate from UDS scope.

This PR is ready for review.

### DIFF
--- a/sil/aeb_gazebo/README.md
+++ b/sil/aeb_gazebo/README.md
@@ -49,15 +49,85 @@ graph LR
 
 | Scenario | Ego speed | Target speed | Initial gap | Description |
 |----------|-----------|--------------|-------------|-------------|
-| `ccrs_20`     | 20 km/h | 0 km/h  | 100 m | Car-to-Car Rear Stationary |
-| `ccrs_30`     | 30 km/h | 0 km/h  | 100 m | |
-| `ccrs_40`     | 40 km/h | 0 km/h  | 100 m | |
-| `ccrs_50`     | 50 km/h | 0 km/h  | 100 m | |
-| `ccrm`        | 50 km/h | 20 km/h |  50 m | Car-to-Car Rear Moving |
-| `ccrb_d2_g12` | 50 km/h | 50 km/h |  12 m | Car-to-Car Rear Braking (decel = 2 m/s²) |
-| `ccrb_d6_g12` | 50 km/h | 50 km/h |  12 m | Car-to-Car Rear Braking (decel = 6 m/s²) |
-| `ccrb_d2_g40` | 50 km/h | 50 km/h |  40 m | Car-to-Car Rear Braking (decel = 2 m/s²) |
-| `ccrb_d6_g40` | 50 km/h | 50 km/h |  40 m | Car-to-Car Rear Braking (decel = 6 m/s²) |
+| `ccrs_20`        | 20 km/h | 0 km/h  | 100 m | Car-to-Car Rear Stationary |
+| `ccrs_30`        | 30 km/h | 0 km/h  | 100 m | |
+| `ccrs_40`        | 40 km/h | 0 km/h  | 100 m | |
+| `ccrs_50`        | 50 km/h | 0 km/h  | 100 m | |
+| `ccrm`           | 50 km/h | 20 km/h |  50 m | Car-to-Car Rear Moving |
+| `ccrb_d2_g12`    | 50 km/h | 50 km/h |  12 m | Car-to-Car Rear Braking (decel = 2 m/s²) |
+| `ccrb_d6_g12`    | 50 km/h | 50 km/h |  12 m | Car-to-Car Rear Braking (decel = 6 m/s²) |
+| `ccrb_d2_g40`    | 50 km/h | 50 km/h |  40 m | Car-to-Car Rear Braking (decel = 2 m/s²) |
+| `ccrb_d6_g40`    | 50 km/h | 50 km/h |  40 m | Car-to-Car Rear Braking (decel = 6 m/s²) |
+| `uds_diagnostic` |  0 km/h | 0 km/h  | 100 m | Stationary harness for the UDS diagnostic panel — no motion, no AEB activity. |
+
+---
+
+## UDS Diagnostics (runtime)
+
+The Zephyr ECU answers UDS services (ISO 14229) over CAN — IDs
+`0x7DF` (request) and `0x7E8` (response) — completing FR-UDS-005
+end-to-end. The SIL exposes those services via HTTP on the `api`
+container and via a side panel in the `dashboard`.
+
+### REST endpoints (`api`, port `:8000`)
+
+| Endpoint | Method | Description |
+|----------|--------|-------------|
+| `/uds/health`          | GET  | Connection state with `canbus` |
+| `/uds/read_did/{did}`  | GET  | `ReadDataByIdentifier` (0x22). Accepts `0xF101` or `61697` |
+| `/uds/snapshot`        | GET  | Reads `0xF100` + `0xF101` + `0xF102` in one call |
+| `/uds/clear_dtc`       | POST | `ClearDiagnosticInformation` (0x14) — clears all DTCs |
+| `/uds/routine_control` | POST | `RoutineControl` (0x31). Body: `{"rid": 769, "value": 0\|1}` |
+
+DIDs supported by the ECU (see `include/aeb_uds.h`):
+
+| DID      | Content              | Decoding |
+|----------|----------------------|----------|
+| `0xF100` | TTC (seconds)        | uint16 little-endian, scale 1/100 |
+| `0xF101` | FSM state            | uint8 raw (0 OFF .. 6 POST_BRAKE) |
+| `0xF102` | Brake pressure (%)   | uint16 little-endian, scale 1/10 |
+
+Known RID: `0x0301` (AEB enable/disable; value `0` disables, `1` enables).
+
+### Dashboard panel
+
+The **UDS** button on the scenario bar (top of the Gazebo iframe)
+opens a side panel that polls `/uds/snapshot` at 5 Hz, displays all
+three DIDs live, and exposes two controls:
+
+- **Clear DTCs (0x14)** — emits `ClearDiagnosticInformation`.
+- **AEB enabled / DISABLED** — toggle via `RoutineControl` on RID `0x0301`.
+
+The panel keeps a colour-coded log of the last 30 actions (green for
+success, red for UDS or HTTP errors).
+
+### Recommended demo flow
+
+```bash
+# After docker compose up, in the dashboard:
+#   1. Pick the 'uds_diagnostic' scenario and click Start.
+#   2. Click the UDS button to open the side panel.
+#   3. Observe FSM = STANDBY and brake = 0%.
+#   4. Click "AEB: enabled" to disable; watch the log row and the
+#      next reads reflect the change.
+```
+
+### Reusable Python client
+
+For external testing or scripting, import the client directly:
+
+```python
+from uds_client import UDSClient, DID_FSM_STATE, RID_AEB_CONTROL
+
+client = UDSClient(host="localhost", port=29536)  # if mapped in compose
+print(client.read_did(DID_FSM_STATE))
+client.routine_control(RID_AEB_CONTROL, value=0)
+```
+
+Unit tests live at
+[`sil/docker/api/tests/test_uds_client.py`](../docker/api/tests/test_uds_client.py)
+and cover encoding/decoding, negative responses, timeouts, and
+concurrent serialisation. Run with `pytest` from `sil/docker/api/`.
 
 ---
 
--- a/sil/aeb_gazebo/config/scenarios.yaml
+++ b/sil/aeb_gazebo/config/scenarios.yaml
@@ -39,6 +39,19 @@ scenarios:
     target_brake_time: 0.0
     pass_criterion: "v_impact < 5 km/h"
 
+  # AEB disabled — control run, demonstrates what happens without intervention.
+  # Identical to ccrs_50 but the api fires UDS RoutineControl(0x0301, 0)
+  # right after the ECU restart, so the FSM stays in OFF the entire run.
+  ccrs_50_no_aeb:
+    name: "CCRs - Ego 50 km/h (AEB DISABLED)"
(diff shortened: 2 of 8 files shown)


RianLinhares (changes requested):
Draft — Review PR #116
Reviewed the nine commits against the SIL architecture, UDS requirements (FR-UDS-001/003/004/005), and the team's portability and independence standards. The Python implementation is well-structured and the test harness is solid, but three hard blockers must be resolved before this exits Draft.

Must Fix
1. Base branch targets feat/sil-gazebo-docker instead of development

All SIL integration flows through development. Merging into feat/sil-gazebo-docker creates an orphaned commit chain — the nine commits here would require a second PR to eventually reach development, and CI gates on development would not fire. Update the base before requesting review.

2. Branch is 111 commits behind development — rebase required

feat/sil-uds-diagnostic has diverged significantly. The 15 unit tests pass against the current tree, but changes already merged upstream are not reflected. A rebase onto current development HEAD is required to surface conflicts before the diff can be reviewed cleanly.

3. No end-to-end live-stack validation

The PR description lists this as an open blocker. The critical path — dashboard panel → /uds/snapshot → scenario_api.py → uds_client.py → TCP-CAN relay (CAN ID 0x7DF) → aeb-ecu container → UDS response (CAN ID 0x7E8) — has not been exercised against a running Docker stack. FakeRelay in test_uds_client.py is the right unit-test pattern but cannot substitute for a live ECU response. A recorded demo (bash sil/start_sil.sh + panel screenshot reading DID 0xF100 live) is required evidence before transitioning to Ready for Review.

Should Fix
4. _send_request() frame layout not independently verified against relay _pack_frame()

uds_client.py packs a 16-byte frame. test_uds_client.py drives FakeRelay, not the real relay, so the byte layout at the socket boundary is untested. A comment citing which field of _pack_frame() maps to which byte offset in _send_request(), or a dedicated integration test, would close this gap. Must be covered in the live-stack demo at minimum.

5. /uds/snapshot serializes three sequential requests through _req_lock

scenario_api.py's /uds/snapshot calls read_did() three times. Since UDSClient serializes all requests via _req_lock, this is three sequential RTTs. At the intended ~5 Hz polling rate from the dashboard this is unlikely to cause visible latency, but a # sequential: 3 × RTT comment would prevent a future maintainer from assuming parallelism.

6. patch_fsm.py present on the branch — out of scope

Acknowledged in the PR notes. Remove or move to a dedicated PR before merge to keep the commit history attributable.

7. .gitignore not updated for Python test artifacts

The PR adds tests/test_uds_client.py driven by pytest. Running it locally generates .pytest_cache/, __pycache__/, and *.pyc under sil/docker/api/tests/. None of these appear in .gitignore. Add them to prevent accidental commits from contributors running tests locally — consistent with the *.gcda/*.gcno pattern already applied on the C side.

8. Stale documentation not addressed

WSL2 native installation instructions and the prototype repo reference remain in README.md unchanged. A # TODO: update annotation would signal to the next contributor that these sections are outdated.

What's Good
FakeRelay test harness — local server without Docker stack is the right isolation level for unit tests. The fixture-per-test pattern ensures no shared state between cases.
15 tests cover adversarial cases — NRC handling, timeout, DID-echo mismatch, and concurrent serialization are all present. The concurrency test with set_auto_reply() per-DID routing demonstrates the lock works correctly under load.
Zero new Python dependencies — uds_client.py uses only stdlib beyond the existing FastAPI image. Consistent with NFR-POR-001 and avoids dependency-management risk in the Docker image.
Lazy client construction with correct HTTP mapping — _get_uds_client() returning HTTP 503 on TCP failure and retrying on next call is the right pattern for a diagnostic path that must not stall the simulation loop. UDSTimeout → 504, UDSError → 502, ValueError → 400 is clean.
start_sil.sh portability fix — removing the hardcoded user path is the correct fix; $(dirname "$0") resolution unblocks any contributor not running on the original machine.
Dead code removal is clean — aeb_node.py, aeb_controller_node.cpp, and docker/canbridge/ removed without leaving stubs or # removed comments. Same pattern as #62.
Panel polling lifecycle — 5 Hz setInterval tears down when the panel closes; zero background traffic during CCRs runs with panel hidden.
Cross-validator independence (ISO 26262-6 §5.4.8) — no .c/.h files under src/, include/, or tests/ were modified. The boundary between embedded C modules and the SIL layer is respected.
Holding approval on Must Fix 1–3. Once the base branch is corrected, the rebase is complete, and live-stack evidence is attached, this is close to mergeable. Items 4–8 can be addressed in the same pass or in a follow-up PR.

RianLinhares (approved):
## PR #116 — `feat/sil-uds-diagnostic` — Review

**Verdict: Approve with comments** — architecture is sound and the
wire-level implementation is correct. One should-fix and two observations
before merge.

---

### What was verified

**Wire protocol correctness**

`uds_client.py` uses 16-byte frames `[4B ID | 1B DLC | 3B pad | 8B data]`
— identical to `can_hal_zephyr.c`. CAN IDs, DIDs, and RID all match the
embedded C layer exactly:

| Constant | Python client | `aeb_uds.c` / `aeb_config.h` |
|---|---|---|
| Request CAN ID | `0x7DF` | matches `aeb_can.c` UDS filter |
| Response CAN ID | `0x7E8` | matches ECU TX ID |
| DID TTC | `0xF100` | `UDS_DID_TTC_VAL` |
| DID FSM state | `0xF101` | `UDS_DID_FSM_STATE_VAL` |
| DID brake press | `0xF102` | `UDS_DID_BRAKE_PRESS_VAL` |
| RID AEB ctrl | `0x0301` | `UDS_RID_AEB_CTRL` |

Scaling also matches: TTC × 100 (uint16 LE), brake × 10 (uint16 LE),
FSM state raw uint8 — all consistent with the `uds_handle_read_did`
encoding in `aeb_uds.c`.

**AEB disable dual-path — correctly motivated**

The PR uses both the ROS topic `/aeb/driver_aeb_enable` → CAN 0x101
AND a one-shot UDS RoutineControl. The inline comment explains why:
UDS RoutineControl persists only until the next `uds_init()` call
(which happens on ECU restart at scenario start). The CAN 0x101 frame
carries the disable bit every cycle, so the override survives ECU
restarts. This is the right approach.

**Unit tests — good coverage**

`test_uds_client.py` uses a `FakeRelay` TCP mock server (not an actual
ECU), testing: wire encoding, response decoding with correct scaling,
negative response handling, timeout paths, DID echo mismatch, and
concurrent serialisation via the threading lock. 15 tests, all
meaningful.

**Dashboard FSM debounce (500 ms)**

Filtering transient OFF flickers on the UI side is correct — the ECU
FSM still transitions correctly; this only prevents confusing visual
noise on the panel. The 500 ms window is reasonable.

**New scenarios**

`ccrs_50_no_aeb` (AEB disabled baseline) and `uds_diagnostic`
(stationary, 100 m gap) are both well-motivated additions.

---

### Should-fix

**Blocking sync UDS calls inside `async def` FastAPI endpoints**

The five UDS endpoints in `scenario_api.py` are defined as `async def`
but call `uds_client.read_did()` / `uds_client.clear_dtc()` etc., which
are **synchronous blocking** socket operations (up to 2 s timeout).
Calling a blocking function directly inside an `async def` endpoint
blocks the entire uvicorn event loop for the duration of the call.
During a 2-second UDS timeout, **all** API endpoints — including
`/scenarios/start` and `/scenarios/stop` — become unresponsive.

Simplest fix: change the UDS endpoints from `async def` to plain `def`.
FastAPI automatically runs sync endpoints in a thread pool, preventing
event loop starvation:

```python
# Before
@app.get("/uds/snapshot")
async def uds_snapshot():
    client = _get_uds_client()
    ...

# After
@app.get("/uds/snapshot")
def uds_snapshot():
    client = _get_uds_client()
    ...
This applies to all five UDS routes: /uds/health, /uds/read_did/{did},

jessica-leoa (approved):

Issue #3, opened 2026-03-02
[DOC] Initialize project repository and README


jessica-leoa:
**Context:**
To kick off the AEB project, we need to establish the repository structure and provide a clear overview of the project, tools, and build instructions.

**Tasks:**
- [x] Create the initial `README.md` with project overview.
- [x] Add CI/CD and Coverage badges.
- [x] Link to the upcoming Wiki pages.

PR #120, opened 2026-05-01
feat(sil): Software-in-the-Loop validation environment (Docker + Zephyr + Gazebo)

renatosfagundes:
# Summary

- Adds a complete Software-in-the-Loop (SIL) validation environment under `sil/` that builds and runs the embedded C code (perception → decision → execution → CAN/UDS) inside a Zephyr `native_sim` ECU, exercised against a Gazebo Classic 11 vehicle simulation.
- Brings up five Docker services with `docker compose up`: TCP CAN relay, Zephyr ECU, Gazebo+ROS 2 (perception + scenario controller + 3D world), FastAPI scenario controller, and an nginx-served web dashboard.
- Web dashboard at `:3000` lets a reviewer pick any Euro NCAP scenario (CCRs / CCRm / CCRb), Start / Stop / Reset, and watch live telemetry (speedometer, FSM badge, TTC bar, brake gauge, alert lamps) alongside the embedded Gazebo 3D view via noVNC.
- Adds CI smoke gate (`.github/workflows/sil.yml`) on PRs touching `sil/`, `src/`, or `include/`: validates the compose file, hadolint on Dockerfiles, builds the small images plus a `--target=zephyr-setup` partial build of the Zephyr image (apt + west update + COPY paths exercised, the ~3-5 min `west build` itself skipped), lints all SIL Python with `ruff` and `compileall`, and `shellcheck`s the entrypoint scripts.
- Heavy images (full Gazebo, full Zephyr `west build`) are not built in CI by design — they exceed the GitHub-hosted runner budget. Documented in `sil/README.md`.
- The Zephyr ECU is built against the **in-tree** C sources (`src/`, `include/`, `stubs/`) via `context: ..` on the `aeb-ecu` service. No `git clone` of a remote branch — the image is reproducible against the exact commit under review.

# Related Issue

- #95 — FSM continues braking below `V_EGO_MIN`. **Fixed in PR #121** (already merged into `development`); this branch was rebased onto the fix and the SIL workaround stack (`patch_fsm.py`, the Dockerfile `RUN python3 …` step) was retired in the same window. The `scenario_controller.py` coast-decel `elif` was kept — Gazebo's planar_move world has no friction, so once AEB releases the ego still needs an explicit deceleration to settle at target speed (independent of #121).
- #42 — Docker-based reproducible development environment. This PR delivers a multi-container SIL setup. Closing #42 may want a slimmer "build the C code in a container" image as well, separate from the SIL.
[...]

Issue #109, opened 2026-04-24
[FIX] Decision: Rule 10.3 fix in aeb_fsm.c cascaded into Rule 10.5 findings


erycaaf:
## Sprint
Sprint 4

## Due Date
2026-04-24

## Objective
The MISRA Rule 10.3 remediation applied in PR #107 used direct casts `(uint8_t)((expr1) || (expr2))` on two Boolean assignments in `fsm_step`. While the cast satisfies Rule 10.3 (explicit conversion to the lvalue's essential type), it introduces Rule 10.5 findings at both sites — "The value of an expression should not be cast to an inappropriate essential type". Rule 10.5 is Advisory by MISRA classification but is not in the `ADVISORY` filter set of `vv-decision.yml`, so the CI MISRA-Required gate would count them and fail once the gate is promoted to blocking.

The idiomatic fix for this combination — Boolean `||` result assigned to `uint8_t` under MISRA C:2012 — is the ternary pattern `(...) ? 1U : 0U`, which satisfies both Rule 10.3 and Rule 10.5 without either cascade.

## Scope
- Refactor `src/decision/aeb_fsm.c:223`:
  ```c
  // Before
  driver_override = (uint8_t)((driver->brake_pedal != 0U) ||
                              (driver->steering_angle > STEERING_OVERRIDE_DEG) ||
                              (driver->steering_angle < -STEERING_OVERRIDE_DEG));

  // After
  driver_override = ((driver->brake_pedal != 0U) ||
                     (driver->steering_angle > STEERING_OVERRIDE_DEG) ||
                     (driver->steering_angle < -STEERING_OVERRIDE_DEG)) ? 1U : 0U;
  ```
- Refactor `src/decision/aeb_fsm.c:242` using the same pattern for `speed_out_of_range`.
- No other Rule 10.3 / 10.5 sites elsewhere in the module scope.

## Acceptance Criteria
- `make misra-decision` reports 0 Rule 10.3 findings and 0 Rule 10.5 findings in Decision scope (the 2 at `aeb_fsm.c:223`, `:242` are removed without reintroducing 10.3).
- `make misra-decision` total findings in scope drop to 8 (down from 10 post-PR-107): Rule 15.5 ×6 + Rule 8.7 ×2, all Advisory already in the gate's `ADVISORY` filter set.
- `make test` — nominal 9/9 passes.
- `make fault-decision` — unchanged semantically (only lvalue form changes, not conditional logic).

[...]

PR #121, opened 2026-05-02
fix(decision): continue braking when v_ego drops below V_EGO_MIN (#95)

### LINKED ISSUE #95: [FIX] Decision (FSM): continue braking when v_ego drops below V_EGO_MIN mid-manoeuvre
## Context

The SIL environment on `feat/sil-gazebo-docker` validates the production C code against Euro NCAP CCR scenarios in a Gazebo + Zephyr `native_sim` + TCP CAN stack. The published CCRs @ 40 km/h result (38.8 km/h → 0 km/h, ego stopped 6.5 m behind the target, collision avoided) **currently depends on a build-time string-replacement patch** of `src/decision/aeb_fsm.c` that ships inside the SIL image (`sil/zephyr_aeb/patch_fsm.py`), not on a proper source change merged into `development`.

Without that patch, the FSM releases the brakes the moment the ego velocity drops below `V_EGO_MIN`, because the "speed out of range → STANDBY" block is reached unconditionally. During a CCRs run the ego is then coasting the last few metres with no braking command, and the Euro NCAP criterion `v_impact < 5 km/h` is not met.

This means the SIL evidence currently published cannot be reproduced against `main` — an auditor running the same C code in any other harness would observe different behaviour. Patching the source at Docker build time is not an acceptable long-term pattern; the fix needs to land in the tree.

## The current (pre-patch) code

In `src/decision/aeb_fsm.c`, in the Priority 3 speed-range block:

```c
    if (speed_out_of_range)
    {
        /* ... prior branches ... */
        else
        {
            fsm_out->fsm_state = (uint8_t)FSM_STANDBY;
        }
        fsm_out->brake_active = 0U;
        fsm_out->alert_level = 0U;
        fsm_out->decel_target = 0.0f;
        fsm_out->state_timer += delta_t_s;
        return;
    }

    /* ===== PRIORITY 4: State Transition Logic ===== */
```

The `else { FSM_STANDBY }` fires for every `speed_out_of_range` case including "actively braking, speed now below `V_EGO_MIN`" — which is the correct case to *continue* braking, not abandon.

## Proposed fix (already prototyped in `patch_fsm.py`)

Distinguish the two sub-cases: if the FSM is currently in a BRAKE_L1/L2/L3/POST_BRAKE state and speed drops below `V_EGO_MIN`, keep braking (let the distance-floor logic in Priority 4 take over); only force STANDBY when not already braking.

```c
    if (speed_out_of_range)
    {
        /* ... prior branches (above-MAX, below-MIN-if-not-braking, etc.) ... */
        else if (current_state < FSM_BRAKE_L1)
        {
            /* Not braking: go to STANDBY */
            fsm_out->fsm_state = (uint8_t)FSM_STANDBY;
            fsm_out->brake_active = 0U;
            fsm_out->alert_level = 0U;
            fsm_out->decel_target = 0.0f;
            fsm_out->state_timer += delta_t_s;
            return;
        }
        /* else: actively braking below V_EGO_MIN — fall through to Priority 4,
         * where distance-floor logic keeps brakes applied until the vehicle
         * fully stops (v < 0.01 → POST_BRAKE). */
    }

    /* ===== PRIORITY 4: State Transition Logic ===== */
```

Note: the patch moves the shared state-update statements (`brake_active = 0`, `alert_level = 0`, `decel_target = 0`, `state_timer`, `return`) **inside** the `else if (current_state < FSM_BRAKE_L1)` block. The fall-through path for "actively braking" must not zero out those fields before Priority 4 runs — Priority 4 will set them.

## Acceptance

- `make test` passes (133+ existing assertions).
- `make mcdc-decision` / `make fault-decision` / `make memory-decision` / `make misra-decision` remain green.
- After merging this fix, rebuild the SIL image **without** `patch_fsm.py` and re-record the CCRs @ 40 km/h result. Expected: 38.8 → 0 km/h, 6.5 m final gap, FSM sequence `STANDBY → WARNING → BRAKE_L1 → BRAKE_L2 → BRAKE_L3 → POST_BRAKE`. Same headline as the patched SIL produces today.
- A new FSM nominal test covering the "actively braking when v < V_EGO_MIN" path (to lock in the behaviour and catch regressions).

## Scope

- **In scope**: `src/decision/aeb_fsm.c` (the Priority 3 speed-range block), a new test in `tests/test_decision.c` or `tests/test_decision_mcdc.c`.
- **Out of scope**: any behavioural change to other FSM states; the TTC layer; perception or CAN layers.

## Why "original author" for this

Per ISO 26262-6 §5.4.8 independence, source fixes to a module are authored by the original module author, not by the cross-validator. The FSM lives under `src/decision/` — Lourenço's module per the strategy doc.

## Follow-up

- Delete `sil/zephyr_aeb/patch_fsm.py` from `feat/sil-gazebo-docker` once this lands.
- Remove the `RUN python3 /app/zephyr_aeb/patch_fsm.py ...` line from `sil/docker/Dockerfile.zephyr`.
- Consolidated SIL V&V report needs an updated evidence section pointing at the post-fix rebuild.

renatosfagundes:
# Summary

- Splits the Priority-3 `speed_out_of_range` branch in `fsm_step` into the three behaviour cases the rest of the FSM already implies: full stop while braking → `POST_BRAKE`, idle out-of-range → `STANDBY`, and the missing case — *actively braking with `v_ego` in `[0.01, V_EGO_MIN)`* now falls through to Priority 4.
- The fall-through path lets `evaluate_desired_state`'s distance-floor ladder (≤5 m → L3, ≤10 m → L2, ≤20 m → L1) keep a `BRAKE_Lx` state and `decel_target > 0` while the ego is still close to the target, so the brake stays applied through the last few metres of every CCRs run.
- Adds two regression tests in `tests/test_decision.c` that lock in the new contract and the unchanged STANDBY-on-idle path.
- Pre-existing behaviour for the `v_ego < 0.01` branch is preserved bit-for-bit (handoff to `POST_BRAKE`, brake released, post_brake_timer reset). `fault_d3_post_brake_exit` exercises this path and still passes.

# Related Issue

Closes #95

# Change Type

- [ ] feat
- [x] fix
- [ ] docs
- [x] test
- [ ] ci
- [ ] refactor/chore

# AEB Areas Affected

- [ ] Perception
- [x] Decision Logic
- [ ] Driver Alerts
- [x] Braking Control
- [x] State Machine
- [ ] CAN Interface
- [ ] UDS Diagnostics
- [ ] Integration
- [ ] Documentation only

# Requirements Impacted

## Functional Requirements

- **FR-DEC-005** — distance-floor logic now actually drives the brake demand all the way to the stop, not just until `v_ego` crosses `V_EGO_MIN`.
- **FR-FSM-001 / 002 / 005** — valid 7-state machine, OFF on disable, POST_BRAKE auto-exit semantics: all unchanged.
- **FR-BRK-001** — brake decel demand stays non-zero through the low-speed approach to a stationary target.

- **NFR-SAF-AVOID** — the residual-speed pass criterion of UNECE R152 / Euro NCAP CCRs is now met by the production code on its own, without a SIL build-time patch.

# Artifacts Updated

- [ ] Requirements document
- [ ] Modeling document
- [x] C source code
- [x] Tests
- [ ] CI workflow
- [ ] CHANGELOG

# Validation Evidence

- [x] Local build passes
- [x] Automated tests pass
- [ ] CI passes (will run on push)
- [x] Traceability updated
- [x] Documentation updated

## Evidence

`make build` (zero-warning gate) and `make test` were run locally on this branch:

```
=== Build OK: zero warnings ===
…
[Decision Logic Tests]
  ✅ FSM enters a BRAKE_Lx state under TTC + distance floor
  ✅ FSM stays in BRAKE_Lx when v_ego dips below V_EGO_MIN while distance floor still applies (#95)
  ✅ Brake remains active when actively braking below V_EGO_MIN
  ✅ decel_target stays > 0 when actively braking below V_EGO_MIN
  ✅ FSM enters POST_BRAKE once v_ego < 0.01 m/s
  ✅ Low-speed idle (not braking) falls to STANDBY
  ✅ Brake is released in STANDBY at low speed
RESULTS: 16 passed, 0 failed

[Integration]
  Results: 26/26 passed, 0 failed
=== All test suites passed ===
```

V&V targets `make mcdc-decision`, `make fault-decision`, `make memory-decision`, `make misra-decision` will run on CI. Local hosts on Windows can't drive `gcc-14` MC/DC coverage, but the per-PR `vv-decision.yml` workflow exercises them.

Affected scenarios: **CCRs at 20 / 30 / 40 / 50 km/h** (target stationary). With this fix the SIL stack on `feat/sil-gazebo-docker` reproduces 38.8 → 0 km/h, ego stopped ~6 m behind target, FSM cascade `STANDBY → WARNING → BRAKE_L1 → BRAKE_L2 → BRAKE_L3 → POST_BRAKE → STANDBY` **without** running `sil/zephyr_aeb/patch_fsm.py`. CCRm and CCRb are unaffected by this code path (target is moving — speed never falls below `V_EGO_MIN` under the original fault-trigger conditions).

# Reviewer Notes

1. The fix is **structurally a hoist**, not a behavioural rewrite of the existing two paths. The `v_ego < 0.01` branch and the not-braking branch each take their own copy of the four shared statements (`brake_active = 0`, `alert_level = 0`, `decel_target = 0`, `state_timer += delta_t_s`) and `return` — exactly as the pre-fix code did when collapsed via the shared trailer. The only new behaviour is the **omission** of those zero-writes on the new fall-through path: Priority 4 now writes them with the correct distance-floor demand instead.

2. Single-exit Advisory (Rule 15.5) count for `fsm_step` rises from 5 → 6. Already in the `ADVISORY` filter set of `vv-decision.yml` per #109, so the MISRA-Required gate stays green. No new Required findings introduced.

3. The new test asserts `fsm_state >= FSM_BRAKE_L1 && fsm_state <= FSM_BRAKE_L3` rather than a specific level — this isolates the fix from the exact distance-floor band the test geometry happens to land in (`distance = 6.5 m` puts us in `≤10 m → BRAKE_L2`, but the test stays robust if a future tuning change re-bands the boundaries).

# Risks / Open Points

1. **POST_BRAKE timer initialisation on the new path.** The existing `case 1` (`v_ego < 0.01`) branch resets `post_brake_timer_s = 0`. The fall-through path doesn't enter POST_BRAKE here — Priority 4 does, on a later cycle when `v_ego` finally crosses `0.01`. The first POST_BRAKE iteration's `state_timer` therefore starts from whatever value the prior BRAKE_Lx left it at (the same as today's pre-fix behaviour for any other transition into POST_BRAKE via Priority 4). No change.

2. **Follow-up on `feat/sil-gazebo-docker`** (PR #120, currently draft): once this fix is merged into `development`, the SIL workaround stack can be retired —
   - delete `sil/zephyr_aeb/patch_fsm.py`,
   - remove the `RUN python3 .../patch_fsm.py ...` step in `sil/docker/Dockerfile.zephyr`,
   - remove the coast-decel branch in `sil/aeb_gazebo/src/scenario_controller.py` (which was compensating for the released-brake behaviour the FSM no longer exhibits).
   I'll open that follow-up against `feat/sil-gazebo-docker` after this PR merges.

3. **Consolidated V&V Report (Decision module).** Section §7.5 desvio D3 should note that `patch_fsm.py` is no longer required after this PR; the SIL evidence is now traceable to the production code in `development`.

### DIFF
--- a/src/decision/aeb_fsm.c
+++ b/src/decision/aeb_fsm.c
@@ -2,7 +2,8 @@
  * @file aeb_fsm.c
  * @brief Implementation of 7-state AEB Finite State Machine.
  * @author Lourenço Jamba Mphili
- * @requirements FR-FSM-001 to FR-FSM-006, FR-DEC-004 to FR-DEC-011
+ * @requirements FR-FSM-001 to FR-FSM-006, FR-DEC-004 to FR-DEC-011,
+ *               FR-BRK-001 (low-speed brake hold via Priority-3 fall-through)
  * @misra Fully compliant with MISRA C:2012
  */
 
@@ -242,10 +243,19 @@ void fsm_step(float32_t delta_t_s,
 
     if (speed_out_of_range != 0U)
     {
-        if ((perception->v_ego < 0.01f) && (current_state >= FSM_BRAKE_L1))
+        /* Three sub-cases. Each terminating branch carries its own copy of
+         * the four trailer statements (brake_active = 0, alert_level = 0,
+         * decel_target = 0, state_timer += delta_t_s) and `return`. The
+         * duplication is intentional: the third sub-case must NOT zero
+         * those fields because Priority 4 will set them itself. Keep the
+         * three blocks in lockstep if any of these trailer fields is ever
+         * extended or renamed. */
+
+        if ((perception->v_ego < V_STOP_THRESHOLD) &&
+            (current_state >= FSM_BRAKE_L1))
         {
             /* Vehicle has come to a full stop while braking — hand off to
-             * POST_BRAKE for the brake-hold timer. */
+             * POST_BRAKE for the brake-hold timer (FR-BRK-005). */
             fsm_out->fsm_state = (uint8_t)FSM_POST_BRAKE;
             m_fsm_mem.post_brake_timer_s = 0.0f;
             fsm_out->brake_active = 0U;
@@ -254,9 +264,14 @@ void fsm_step(float32_t delta_t_s,
             fsm_out->state_timer += delta_t_s;
             return;
         }
-        if (current_state < FSM_BRAKE_L1)
+        if ((current_state < FSM_BRAKE_L1) || (current_state > FSM_BRAKE_L3))
         {
-            /* Not braking and out-of-range: drop cleanly to STANDBY. */
+            /* Not in an active braking state (STANDBY / WARNING / OFF /
+             * POST_BRAKE) and out-of-range: drop cleanly to STANDBY. The
+             * POST_BRAKE case is included here on purpose — its hold
+             * timer must not be lengthened by a low-speed dip; if v_ego
+             * has not yet fallen below V_STOP_THRESHOLD we treat it as
+             * an out-of-range exit. */
             fsm_out->fsm_state = (uint8_t)FSM_STANDBY;
             fsm_out->brake_active = 0U;
             fsm_out->alert_level = 0U;
@@ -264,11 +279,12 @@ void fsm_step(float32_t delta_t_s,
             fsm_out->state_timer += delta_t_s;
             return;
         }
-        /* Actively braking with v_ego in [0.01, V_EGO_MIN): fall through
-         * to Priority 4. The distance-floor logic in evaluate_desired_state
-         * resolves to BRAKE_L1/L2/L3 while the ego is still close to the
-         * target, so the brake stays applied until v_ego < 0.01 hits the
-         * branch above and hands off to POST_BRAKE.
+        /* Actively braking (BRAKE_L1/L2/L3) with v_ego in
+         * [V_STOP_THRESHOLD, V_EGO_MIN): fall through to Priority 4. The
+         * distance-floor logic in evaluate_desired_state resolves to
+         * BRAKE_L1/L2/L3 while the ego is still close to the target, so
+         * the brake stays applied until v_ego < V_STOP_THRESHOLD hits
+         * the first branch above and hands off to POST_BRAKE.
          *
          * Closes #95: previously every speed_out_of_range case forced
          * STANDBY and zeroed the brake — the ego coasted the last few
--- a/src/decision/aeb_fsm.c
+++ b/src/decision/aeb_fsm.c
@@ -244,18 +244,39 @@ void fsm_step(float32_t delta_t_s,
     {
         if ((perception->v_ego < 0.01f) && (current_state >= FSM_BRAKE_L1))
         {
+            /* Vehicle has come to a full stop while braking — hand off to
+             * POST_BRAKE for the brake-hold timer. */
             fsm_out->fsm_state = (uint8_t)FSM_POST_BRAKE;
             m_fsm_mem.post_brake_timer_s = 0.0f;
+            fsm_out->brake_active = 0U;
+            fsm_out->alert_level = 0U;
+            fsm_out->decel_target = 0.0f;
+            fsm_out->state_timer += delta_t_s;
+            return;
         }
-        else
+        if (current_state < FSM_BRAKE_L1)
         {
+            /* Not braking and out-of-range: drop cleanly to STANDBY. */
             fsm_out->fsm_state = (uint8_t)FSM_STANDBY;
+            fsm_out->brake_active = 0U;
+            fsm_out->alert_level = 0U;
+            fsm_out->decel_target = 0.0f;
+            fsm_out->state_timer += delta_t_s;
+            return;
         }
-        fsm_out->brake_active = 0U;
-        fsm_out->alert_level = 0U;
-        fsm_out->decel_target = 0.0f;
-        fsm_out->state_timer += delta_t_s;
-        return;
+        /* Actively braking with v_ego in [0.01, V_EGO_MIN): fall through
+         * to Priority 4. The distance-floor logic in evaluate_desired_state
+         * resolves to BRAKE_L1/L2/L3 while the ego is still close to the
+         * target, so the brake stays applied until v_ego < 0.01 hits the
+         * branch above and hands off to POST_BRAKE.
+         *
+         * Closes #95: previously every speed_out_of_range case forced
+         * STANDBY and zeroed the brake — the ego coasted the last few
+         * metres of every CCRs run with no braking command, missing the
+         * Euro NCAP residual-speed criterion. The SIL stack worked
+         * around this with a build-time string patch
+         * (sil/zephyr_aeb/patch_fsm.py); with this change the patch can
+         * be removed. */
     }
 
     /* ===== PRIORITY 4: State Transition Logic ===== */
--- a/src/decision/aeb_fsm.c
+++ b/src/decision/aeb_fsm.c
@@ -284,15 +284,7 @@ void fsm_step(float32_t delta_t_s,
          * distance-floor logic in evaluate_desired_state resolves to
          * BRAKE_L1/L2/L3 while the ego is still close to the target, so
          * the brake stays applied until v_ego < V_STOP_THRESHOLD hits
-         * the first branch above and hands off to POST_BRAKE.
-         *
-         * Closes #95: previously every speed_out_of_range case forced
-         * STANDBY and zeroed the brake — the ego coasted the last few
-         * metres of every CCRs run with no braking command, missing the
-         * Euro NCAP residual-speed criterion. The SIL stack worked
-         * around this with a build-time string patch
-         * (sil/zephyr_aeb/patch_fsm.py); with this change the patch can
-         * be removed. */
+         * the first branch above and hands off to POST_BRAKE. */
     }
 
     /* ===== PRIORITY 4: State Transition Logic ===== */
--- a/tests/test_decision.c
+++ b/tests/test_decision.c
@@ -193,12 +193,13 @@ static void test_fsm_continue_braking_below_vmin(void)
     {
         fsm_step(0.01f, &perc, &driver, &ttc, &fsm_out);
     }
-    TEST_ASSERT(fsm_out.fsm_state >= (uint8_t)FSM_BRAKE_L1,
+    TEST_ASSERT(fsm_out.fsm_state >= (uint8_t)FSM_BRAKE_L1 &&
+                fsm_out.fsm_state <= (uint8_t)FSM_BRAKE_L3,
                 "FSM enters a BRAKE_Lx state under TTC + distance floor");
 
-    /* Now ego drops below V_EGO_MIN (10 km/h = 2.78 m/s) but is not yet
-     * fully stopped. Target is still close (distance still in floor). */
-    perc.v_ego = 1.5f;            /* < V_EGO_MIN, > 0.01 */
+    /* Now ego drops below V_EGO_MIN (10 km/h = 2.78 m/s) but is still
+     * above V_STOP_THRESHOLD. Target is still close (distance in floor). */
+    perc.v_ego = 1.5f;            /* < V_EGO_MIN, > V_STOP_THRESHOLD */
     perc.distance = 6.5f;         /* still under the 10 m floor */
     fsm_step(0.01f, &perc, &driver, &ttc, &fsm_out);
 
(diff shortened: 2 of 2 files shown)


RianLinhares (changes requested):
## PR Review — `fix(decision): continue braking when v_ego drops below V_EGO_MIN (#95)`

**Decision: Request Changes** — one Must Fix before merge.

---

### Must Fix

**1. `current_state >= FSM_BRAKE_L1` silently changes `POST_BRAKE` behavior**

`FSM_POST_BRAKE` has enum value 6, which is `>= FSM_BRAKE_L1` (3). This means when `speed_out_of_range` is true **and** `v_ego` is in `[0.01, V_EGO_MIN)` **and** `current_state == FSM_POST_BRAKE`, the new code falls through to Priority 4 instead of the original behavior (forced STANDBY + brake zeroed).

This behavioral change may be correct — it makes sense for the POST_BRAKE timer to run to completion even at low speed. However:

- It is not documented in the PR summary or Reviewer Notes (the PR describes the fix for `BRAKE_L1/L2/L3` only, with no mention of POST_BRAKE)
- No test covers this path

Two options:
- **If intentional**: document it explicitly and add a `test_fsm_post_brake_low_speed_completes_timer` test confirming the timer runs normally when `v_ego` is in `[0.01, V_EGO_MIN)` during POST_BRAKE.
- **If unintentional**: restrict the fall-through condition to exclude POST_BRAKE:

```c
if (current_state >= FSM_BRAKE_L1 && current_state <= FSM_BRAKE_L3)
{
    /* fall through to Priority 4 */
}
```

---

### Should Fix

**2. Code duplication in the two returning branches**

The four shared lines:

```c
fsm_out->brake_active = 0U;
fsm_out->alert_level  = 0U;
fsm_out->decel_target = 0.0f;
fsm_out->state_timer += delta_t_s;
return;
```

were written once in the original code after the `if/else`. The hoist duplicated them into two independent copies. This increases the risk of future divergence if one branch is updated and the other is not. Consider extracting to a `fsm_exit_safe()` helper, or at minimum add a comment explicitly stating the duplication is intentional.

**3. `0.01f` is a magic number that now appears twice**

The full-stop threshold (`v_ego < 0.01f`) now appears in the branch condition. Define it as a named constant in `aeb_config.h` — e.g. `V_EGO_STOP_MS` — to make the intent explicit and prevent drift between the two occurrences.

**4. Doxygen `@requirements` header not updated**

The PR body lists `FR-DEC-005` and `FR-BRK-001` as requirements now satisfied by this fix. The `@requirements\` doxygen tag in `aeb_fsm.c` should be updated to include them.

---

### What's Good

- The fix is minimal and surgical — only the Priority-3 guard changes; `evaluate_desired_state()` in Priority 4 is untouched.
- The three-case decomposition with descriptive inline comments makes the intent immediately clear to any future reader.
- Regression tests cover the bug-fix path (brake held through low-speed approach), the unchanged STANDBY path (idle out-of-range), and the `v_ego < 0.01` handoff to POST_BRAKE.
- The PR body explicitly states that `v_ego < 0.01` behavior is preserved bit-for-bit — important for reviewing without concern about regressions in the full-stop path.
- The Rule 15.5 exit-count increase (5 → 6) is acknowledged and already in the ADVISORY filter set per #109 — no MISRA gate surprise.
- Follow-up clearly identified: retire `patch_fsm.py` from the SIL stack (PR #120) after merge.
- FR/NFR fields populated with specific IDs, not left blank.

---

**Summary:** The core fix is correct and well-described. The blocker is the undocumented and untested behavioral change for `POST_BRAKE` — one sentence in the PR body and one test resolve it. The Should Fix items are all code quality concerns and do not affect runtime correctness.

renatosfagundes:
@RianLinhares — addressed in `063a297`. Quick walk-through of each item:

**Must Fix #1 (POST_BRAKE)** — went with the "restrict" option. Guard is now `(current_state < FSM_BRAKE_L1) || (current_state > FSM_BRAKE_L3)`, so the fall-through fires only for `BRAKE_L1/L2/L3`. POST_BRAKE keeps its original out-of-range behaviour (forced STANDBY). No new test needed — the existing regression test sets up `BRAKE_L2`, which is in the tightened range, so it still exercises exactly the fall-through path.

**Should Fix #2 (duplication)** — added an explicit comment at the top of the `speed_out_of_range` block noting that the four-line trailer (`brake_active = 0`, `alert_level = 0`, `decel_target = 0`, `state_timer += delta_t_s`, `return`) is intentionally duplicated, with a directive to keep them in lockstep on future changes. Didn't extract to a helper — that would have to also update `state_timer` consistently across the call sites and is more invasive than the duplication risk warrants here.

**Should Fix #3 (`0.01f` magic number)** — happily resolved by the existing `V_STOP_THRESHOLD` macro in `aeb_config.h:50` (already named exactly this concept). No new constant needed.

**Should Fix #4 (`@requirements` header)** — added `FR-BRK-001 (low-speed brake hold via Priority-3 fall-through)` to the `aeb_fsm.c` doxygen tag.

All three local gates still green: `make build` zero-warning, `make test` 26/26, `test_decision` 16/16.

RianLinhares (dismissed):
## PR Review (follow-up) — `fix(decision): continue braking when v_ego drops below V_EGO_MIN (#95)`

**Decision: Approve** — all items from the previous review addressed in `063a297`.

---

### Items Resolved

**Must Fix #1 (POST_BRAKE fall-through)** — resolved with the "restrict" option. The new guard:

```c
if ((current_state < FSM_BRAKE_L1) || (current_state > FSM_BRAKE_L3))
```

correctly excludes `POST_BRAKE` (enum 6 > `FSM_BRAKE_L3` = 5) from the fall-through path and forces it to STANDBY. The inline comment explicitly states *why* POST_BRAKE is excluded — not just how — which is exactly what this case needed.

**Should Fix #2 (duplication)** — the lockstep comment at the top of the `speed_out_of_range` block clearly documents the intent and gives a maintenance directive. Acceptable resolution without the invasiveness of a helper function.

**Should Fix #3 (`0.01f` magic number)** — replaced with the pre-existing `V_STOP_THRESHOLD` macro. Clean resolution.

**Should Fix #4 (`@requirements` header)** — `FR-BRK-001` added to the doxygen tag with a short description of the mechanism. Traceable.

---

### One Observation (non-blocking)

The first branch still uses `current_state >= FSM_BRAKE_L1`, which includes `POST_BRAKE` (6). This means `POST_BRAKE + v_ego < V_STOP_THRESHOLD` re-enters `POST_BRAKE` with `post_brake_timer_s` reset to zero. This is pre-existing behaviour from the original code and is not introduced by this PR — but it implies the vehicle will never exit POST_BRAKE while it remains fully stopped (timer resets every cycle). Worth a follow-up issue if that edge case has not been analysed.

---

### Final State

All three local gates green (`make build` zero-warning, `make test` 26/26, `test_decision` 16/16). The fix is correct, minimal, and well-documented. Ready to merge after CI confirms.

erycaaf:
# Review — PR #121
The core fix is correct and the blocker from the prior review (POST_BRAKE fall-through) was resolved precisely. Branch is fresh: merge-base = af5a4ec, current development tip. Three quality items in the tests and documentation were not caught by the prior review.
## Should Fix
### 1. Warm-up assertion lacks upper bound in test

test_fsm_continue_braking_below_vmin, line 42:
TEST_ASSERT(fsm_out.fsm_state >= (uint8_t)FSM_BRAKE_L1, ...);

The check has only a lower bound; it would pass silently if the 100-cycle warm-up ended in FSM_POST_BRAKE (value 6 > 3). The critical assertion immediately below is correctly bounded (>= BRAKE_L1 && <= BRAKE_L3); the warm-up check should follow the same pattern:

TEST_ASSERT(fsm_out.fsm_state >= (uint8_t)FSM_BRAKE_L1 &&
            fsm_out.fsm_state <= (uint8_t)FSM_BRAKE_L3, ...);

### 2. Magic number 0.005f survives in the test
V_STOP_THRESHOLD was correctly introduced in production code but is not included in test_decision.c. Line 62 uses perc.v_ego = 0.005f; lines 47, 60, and the assertion message on line 65 still reference 0.01. If the constant changes in aeb_config.h, the test silently diverges. Replace with:

perc.v_ego = V_STOP_THRESHOLD * 0.5f;   /* guaranteed below the threshold */
and update comments/messages to read < V_STOP_THRESHOLD instead of < 0.01.

### 3. Historical "Closes #95" paragraph in production code comment
The third sub-case in aeb_fsm.c ends with an 8-line block referencing the issue number and describing the SIL stack patch_fsm.py workaround; historical context that belongs in the commit message and PR body, not in inline code. The mechanism description in the first five lines is sufficient. Remove from "Closes #95: previously..." onwards.

renatosfagundes:
@erycaaf — addressed in `a68f626`. Quick walk-through:

**1. Warm-up assertion bound** — tightened to `>= FSM_BRAKE_L1 && <= FSM_BRAKE_L3` so it mirrors the critical assertion two lines below. Now catches a stray POST_BRAKE landing during the 100-cycle warm-up.

**2. Magic numbers in the test** — replaced `0.005f` with `V_STOP_THRESHOLD * 0.5f`, updated the comment that said "< 0.01" to "< V_STOP_THRESHOLD", and updated the assertion message accordingly. `V_STOP_THRESHOLD` reaches the test transitively via `aeb_fsm.h` → `aeb_config.h`, so no additional include was needed.

**3. Historical "Closes #95" paragraph in `aeb_fsm.c`** — trimmed the 8-line block that described the pre-#95 behaviour and the SIL `patch_fsm.py` workaround. The 5-line mechanism comment stays; the historical context lives in the commit message and the PR body where it belongs.

Local gates still green: `make build` zero-warning, `test_decision` 16/16, full `make test` 26/26.

erycaaf (approved):

RianLinhares (approved):
LGTM

PR #126, opened 2026-05-10
ci(vv): aggregate Pages publish into one workflow to avoid concurrency cancellations

### LINKED ISSUE #127: Pages publish race cancels 2-3 V&V reports per push
## Problem

On every push to `development` or `main`, 2-3 of the five module V&V workflows fail to publish their HTML report to GitHub Pages even though the V&V gates themselves all pass. The reports on Pages stay stuck on older deploys, breaking the link between the latest commit on the trunk branches and the published evidence.

## Reproduction

Latest example — push of `5cf8eba` to `main` (run set https://github.com/erycaaf/AEB-stellantis-project/actions, 2026-05-06):

| Workflow              | Stack job | publish-pages |
|-----------------------|-----------|---------------|
| Perception V&V        | ✅        | ✅ deployed   |
| PID + Alert V&V       | ✅        | ✅ deployed   |
| Publish V&V index page| —         | ✅ deployed   |
| **CAN V&V**           | ✅        | **❌ cancelled** |
| **Decision V&V**      | ✅        | **❌ cancelled** |
| **UDS V&V**           | ✅        | **❌ cancelled** |

Inspecting any of the cancelled `publish-pages` jobs shows them entering "completed/cancelled" state ~3 s after start with **0 steps executed**, i.e., the runner never picked them up.

## Root cause

Every per-module workflow (`vv-can.yml`, `vv-decision.yml`, `vv-perception.yml`, `vv-pid-alert.yml`, `vv-uds.yml`) and `publish-index.yml` declares the **same** concurrency group on its publish step:

```yaml
concurrency:
  group: gh-pages-deploy
  cancel-in-progress: false
```

GitHub Actions enforces a strict policy on `cancel-in-progress: false` groups: *at most one running + one pending* in the group at any time. Any third arrival to the queue **cancels the previously pending job**. With 5 V&V workflows + the index workflow all triggering in parallel on the same push, three of the five module publishes lose the race and never run.

The V&V gates themselves are unaffected — only the deploy step is cancelled. This is a Pages-publication problem, not a verification problem.

## Expected behaviour

Every push to `development`/`main` that produces a successful V&V run for **all five** modules should publish *all five* module reports to gh-pages, in the same `vv_<module>/<branch>/` subdirectory hierarchy already in use today.

## Acceptance criteria

- A push to `development` or `main` produces exactly **one Pages deploy commit per push**, containing the latest HTML for every module whose stack job passed.
- Modules whose stack job fails on the push keep their last successful Pages deploy intact (no regression on `keep_files: true` semantics).
- No `publish-pages` jobs reported as cancelled by GitHub Actions on subsequent runs.

## Related

- The `publish-index.yml` workflow that deploys `docs/index.html` should keep working unchanged. The fix scope is the per-module reports under `vv_<module>/`.
- See discussion of `cancel-in-progress` semantics in [GitHub Actions docs](https://docs.github.com/en/actions/using-jobs/using-concurrency).

renatosfagundes:
# Summary
- Replace the per-workflow `publish-pages` jobs with a single aggregator workflow (`.github/workflows/publish-vv-reports.yml`) triggered by `workflow_run` on completion of any of the five module V&V workflows.
- The aggregator gates on the `gh run list` API for all five sibling workflows on the same SHA; only the last completion event proceeds to deploy, earlier events exit silently.
- One Pages deploy per push, containing every module's latest HTML report — no more 2-of-5 cancelled publishes from the `gh-pages-deploy` concurrency-group race.
- Deletes the `publish-pages` job from `vv-can.yml`, `vv-decision.yml`, `vv-perception.yml`, `vv-pid-alert.yml`, and `vv-uds.yml` (sibling per-module artefact uploads stay; the aggregator pulls them via `gh run download`).
- `publish-index.yml` (root `docs/index.html` landing page) is untouched and continues to deploy independently.
- Skips deploy when any of the five V&V runs ends in failure for the SHA, so partial pushes never overwrite the last green Pages snapshot (preserves `keep_files: true` semantics for unaffected modules).

# Related Issue
Closes #127

# Change Type
- [ ] feat
- [ ] fix
- [ ] docs
- [ ] test
- [x] ci
- [ ] refactor/chore

# AEB Areas Affected
- [ ] Perception
- [ ] Decision Logic
- [ ] Driver Alerts
- [ ] Braking Control
- [ ] State Machine
- [ ] CAN Interface
- [ ] UDS Diagnostics
- [ ] Integration
- [x] Documentation only

# Requirements Impacted
## Functional Requirements
- N/A — no module behaviour changes; CI publication path only.

- NFR-VV-CI: every push to `development`/`main` whose V&V gates pass must publish all five module reports to GitHub Pages.

# Artifacts Updated
- [ ] Requirements document
- [ ] Modeling document
- [ ] C source code
- [ ] Tests
- [x] CI workflow
- [ ] CHANGELOG

# Validation Evidence
- [x] Local build passes
- [ ] Automated tests pass
- [ ] CI passes
- [ ] Traceability updated
- [ ] Documentation updated

## Evidence

**Failure observed (issue #127, latest example — push `5cf8eba` on `main`):**

| Workflow              | Stack job | publish-pages |
|-----------------------|-----------|---------------|
| Perception V&V        | success   | deployed      |
| PID + Alert V&V       | success   | deployed      |
| Publish V&V index     | —         | deployed      |
| **CAN V&V**           | success   | **cancelled** |
| **Decision V&V**      | success   | **cancelled** |
| **UDS V&V**           | success   | **cancelled** |

Cancelled `publish-pages` jobs entered `completed/cancelled` ~3 s after start with `steps: []` — the runner never picked them up, classic "third arrival to a `cancel-in-progress: false` group cancels the previously pending job" pattern.

**Local validation of the new aggregator:**

```bash
$ python -c "import yaml; yaml.safe_load(open('.github/workflows/publish-vv-reports.yml'))"
$ for f in vv-can vv-decision vv-perception vv-pid-alert vv-uds; do \
    python -c "import yaml; yaml.safe_load(open('.github/workflows/${f}.yml'))"; \
  done
# all parse clean — no orphan comment blocks left after removing publish-pages
```

**Gate logic check (manual trace):**

The aggregator fires on each of the 5 V&V completions. For the same SHA, gate output `go=true` only when `gh run list --workflow <wf> --commit <SHA>` returns `status=completed` for all 5 *and* `conclusion=success` for all 5. First 4 firings see `all_done=false` and exit silently; the 5th completes the fan-in and deploys once.

**Non-regression on partial failures:**

If a stack job fails for one module, that workflow's run completes with `conclusion=failure`; the gate sets `go=false` and the deploy is skipped. The previous Pages snapshot stays in place (`keep_files: true` semantics), matching the third acceptance criterion in the issue.

# Reviewer Notes

- Focus on the `Gate` step in `publish-vv-reports.yml`: it's the only piece preventing duplicate deploys. The "exit silently when not all 5 done" branch must not error or it cancels its own concurrency slot.
- `gh run download <run_id> -n <artefact_name>` relies on the existing artefact-naming convention `vv-<module>-reports-run<N>` (uploads in each `vv-<module>.yml`). Worth a sanity check that I removed only the `publish-pages` block and not the `actions/upload-artifact` step.
- The aggregator still keeps `concurrency: gh-pages-deploy` so it serializes against `publish-index.yml`. Two deploys per push (this one + index) is fine; the race only triggered above 2 in flight.

# Risks / Open Points

- **Residual concurrency risk:** any future workflow that joins the `gh-pages-deploy` group will reintroduce the 1+1 cap. We should gate new Pages-deploying workflows behind the aggregator pattern by default.
- **Artefact retention:** `gh run download` reads from per-run artefacts which expire after the repo's default retention (90 d). A re-run of *the aggregator alone* on an old SHA would fail to fetch artefacts; this is acceptable because Pages publication is a forward-only operation tied to the latest push.
- **Manual recovery on partial failures:** if 4 of 5 modules pass and 1 fails, no Pages update happens this push. After the failing module is re-run and succeeds, the user must re-trigger one of the other 4 V&V workflows (or push a no-op) to fire the aggregator again. Documenting this in the README is a follow-up.
- **No automated test for the gate logic** — relies on the next push to `development` to verify in production CI.

### DIFF
--- a/.github/workflows/publish-index.yml
+++ b/.github/workflows/publish-index.yml
@@ -1,10 +1,10 @@
 name: Publish V&V index page
 
 # Deploys docs/index.html to the root of the gh-pages branch whenever it
-# changes on development or main. Per-module workflows (vv-uds.yml,
-# vv-pid.yml, etc.) deploy their HTML reports to subdirectories of the
-# same gh-pages branch using keep_files: true, so this index page sits
-# at the root and links into them.
+# changes on development or main. The per-module HTML reports are
+# published by publish-vv-reports.yml into subdirectories of the same
+# gh-pages branch (vv_<module>/<branch>/), using keep_files: true so
+# this index page sits at the root and links into them.
 
 on:
   push:
--- a/.github/workflows/publish-vv-reports.yml
+++ b/.github/workflows/publish-vv-reports.yml
@@ -1,0 +1,151 @@
+name: Publish V&V reports
+
+# Aggregator that publishes the per-module HTML reports to GitHub Pages
+# in a single deploy after ALL five module V&V workflows have completed
+# for the same commit. Prior design ran a publish-pages job inside each
+# of the V&V workflows; with five concurrent push triggers, GitHub
+# Actions' "1 running + 1 queued in concurrency group" rule cancelled
+# 3 of 5 publish jobs, leaving stale reports on Pages.
+#
+# This workflow listens for completion of any of the five V&V workflows.
+# Each invocation checks whether all five sibling workflows have
+# completed successfully on the same SHA. The last one to complete
+# performs the deploy; earlier invocations exit silently. Net result:
+# exactly one Pages deploy per push, with all five module artefacts.
+#
+# The publish-index.yml workflow continues to deploy the root
+# docs/index.html landing page; this aggregator deploys the per-module
+# subdirectories under it.
+
+on:
+  workflow_run:
+    workflows:
+      - "CAN V&V"
+      - "Decision V&V"
+      - "Perception V&V"
+      - "PID + Alert V&V"
+      - "UDS V&V"
+    types: [completed]
+
+permissions:
+  contents: write
+  actions: read
+
+concurrency:
+  group: gh-pages-deploy
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    name: Aggregate and publish to gh-pages
+    runs-on: ubuntu-24.04
+    if: |
+      github.event.workflow_run.conclusion == 'success' &&
+      (github.event.workflow_run.head_branch == 'development' ||
+       github.event.workflow_run.head_branch == 'main')
+    steps:
+      - name: Gate — wait until all 5 V&V workflows complete for this SHA
+        id: gate
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SHA: ${{ github.event.workflow_run.head_sha }}
+        run: |
+          set -e
+          workflows=("CAN V&V" "Decision V&V" "Perception V&V" "PID + Alert V&V" "UDS V&V")
+          all_done=true
+          all_success=true
+          for wf in "${workflows[@]}"; do
+            row=$(gh run list \
+                --workflow "$wf" \
+                --commit "$SHA" \
+                --limit 1 \
+                --json status,conclusion,databaseId \
+                --jq '.[0] // {}')
+            status=$(echo "$row"     | jq -r '.status     // "unknown"')
+            concl=$(echo  "$row"     | jq -r '.conclusion // ""')
+            id=$(echo     "$row"     | jq -r '.databaseId // 0')
+            echo "  $wf: status=$status, conclusion=$concl, run=$id"
+            if [ "$status" != "completed" ]; then
+              all_done=false
+            elif [ "$concl" != "success" ]; then
+              # Completed but not successful — don't deploy, but don't keep waiting either.
+              all_success=false
+            fi
+          done
+
+          if [ "$all_done" != "true" ]; then
+            echo "Not all sibling workflows have completed yet. Exiting silently."
+            echo "go=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          if [ "$all_success" != "true" ]; then
+            echo "All workflows completed but at least one failed. Skipping deploy so reports stay aligned."
+            echo "go=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          echo "All 5 V&V workflows complete & successful for SHA $SHA — proceeding with deploy."
+          echo "go=true" >> "$GITHUB_OUTPUT"
+
+      - name: Stage site directory
+        if: steps.gate.outputs.go == 'true'
+        run: mkdir -p site
+
+      - name: Download artefacts from each V&V run
+        if: steps.gate.outputs.go == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SHA: ${{ github.event.workflow_run.head_sha }}
+        run: |
+          set -e
+          # Map: workflow name → site subdirectory + artefact-name prefix
+          declare -A site_dir=(
+            ["CAN V&V"]="vv_can"
+            ["Decision V&V"]="vv_decision"
+            ["Perception V&V"]="vv_perception"
+            ["PID + Alert V&V"]="vv_pid_alert"
+            ["UDS V&V"]="vv_uds"
+          )
+          declare -A art_prefix=(
+            ["CAN V&V"]="vv-can-reports-run"
+            ["Decision V&V"]="vv-decision-reports-run"
+            ["Perception V&V"]="vv-perception-reports-run"
+            ["PID + Alert V&V"]="vv-pid-alert-reports-run"
+            ["UDS V&V"]="vv-uds-reports-run"
+          )
+
+          for wf in "${!site_dir[@]}"; do
+            module="${site_dir[$wf]}"
+            prefix="${art_prefix[$wf]}"
+
+            row=$(gh run list \
+                --workflow "$wf" \
+                --commit "$SHA" \
+                --limit 1 \
+                --json databaseId,number \
+                --jq '.[0]')
+            run_id=$(echo "$row"     | jq -r '.databaseId')
+            run_number=$(echo "$row" | jq -r '.number')
+
+            artefact_name="${prefix}${run_number}"
+            target="site/${module}"
+            mkdir -p "$target"
+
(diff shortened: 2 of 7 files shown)


RianLinhares (changes requested):
## Review

**Verdict: Changes requested** — one silent deadlock and one broken URL family, both fixable.

---

### What this PR is trying to do — and why the goal is right

The old design gave every V&V workflow its own `publish-pages` job. With five concurrent pushes on the same commit, GitHub's "1 running + 1 queued" concurrency rule cancels three of the five publish jobs, leaving Pages with a partial snapshot. Collapsing publishing into a single aggregator is the correct architectural fix.

---

### Bug 1 — Aggregator deadlocks on any single-module push (HIGH)

Each V&V workflow has a narrow `paths:` filter. A commit that only touches `src/communication/aeb_can.c` triggers `CAN V&V` only; the other four workflows never run for that SHA.

The gate loop queries `gh run list --workflow "$wf" --commit "$SHA" --limit 1` for all five workflows. For the four that did not run, that command returns `[]`, `row` becomes `{}`, and `status` evaluates to `"unknown"`. That sets `all_done=false`, so the aggregator exits silently with `go=false`. **Pages is never updated.**

The only commit that would trigger a successful deploy is one that simultaneously touches the path filters for all five modules — an event that essentially never happens in normal development.

**Fix:** treat "no run found for this SHA" as "not relevant to this push":

    if [ "$id" = "0" ] || [ -z "$id" ]; then
      echo "    $wf: no run for this SHA — skipping"
      continue   # not applicable, don't block all_done
    fi
    if [ "$status" != "completed" ]; then
      all_done=false
    elif [ "$concl" != "success" ]; then
      all_success=false
    fi

And in the download loop:

    if [ "$run_id" = "null" ] || [ "$run_id" = "0" ]; then
      echo "  $wf: no run for this SHA — skipping artifact"
      continue
    fi

With this fix the aggregator deploys as soon as every *triggered* workflow for that SHA succeeds, and silently skips modules whose path filters did not match. This preserves the old per-module behaviour while eliminating the concurrency race.

---

### Bug 2 — Step summary browse-links become 404s after this PR (MEDIUM)

The old per-module `publish-pages` jobs used `destination_dir: vv_can/${{ github.ref_name }}`, so the live URL was `…/vv_can/development/…`. The new aggregator has no `destination_dir` and downloads artifacts to `site/vv_can/`, which lands at `vv_can/` — **the branch segment is gone.**

Every V&V workflow still emits this line in its step summary:

    📊 Browse reports: https://erycaaf.github.io/AEB-stellantis-project/vv_can/${{ github.ref_name }}/

That URL resolves to the old location (`vv_can/development/`), which no longer exists.

Two ways to fix this:

**Option A** — drop branch namespacing (simpler): remove the `/${{ github.ref_name }}` suffix from the step summary browse-links in all five V&V workflow files.

**Option B** — preserve branch namespacing in the aggregator: change the download targets in `publish-vv-reports.yml` to `site/${module}/${{ github.event.workflow_run.head_branch }}` — no change needed in the step summary links.

---

### Minor — stale header comments in the five V&V workflows (LOW)

Each V&V workflow header still reads:

    # Reports are published in two layers:
    #   2. GitHub Pages (gh-pages branch) on push to development/main —
    #      navigable HTML reports under https://…/vv_<module>/<branch>/.

This is now inaccurate. Update to:

    #   2. GitHub Pages — published by publish-vv-reports.yml once all five
    #      module V&V workflows complete successfully for the same commit.

---

### What looks good

- Concurrency group `gh-pages-deploy` with `cancel-in-progress: false` correctly serialises the aggregator against `publish-index.yml`. ✓
- `workflow_run` branch guard (`head_branch == 'development' || 'main'`) correctly limits deploys to the two stable branches. ✓
- Graceful download failure prevents a missing artifact from aborting the whole deploy. ✓
- Commit message includes branch and SHA, making the gh-pages history auditable. ✓
- The five `publish-pages` removals are clean — no leftover `permissions: contents: write` that would expose unnecessary scope. ✓

The architecture is sound; the two bugs above are the only blockers to merge.

RianLinhares (approved):
**Verdict: Approve** — both blockers are resolved cleanly; the aggregator now works correctly for real-world pushes.

---

### Bug 1 — Fixed ✓

The gate loop now skips workflows that produced no run for the SHA instead of setting `all_done=false`:

    if [ "$id" = "0" ] || [ "$id" = "null" ] || [ -z "$id" ]; then
      echo "  $wf: no run for this SHA (path filter did not match) — skipping"
      continue
    fi

The `ran_count` guard also handles the degenerate case where no module ran at all (`go=false`, silent exit). The log line `"$ran_count triggered V&V workflow(s) complete & successful for SHA $SHA"` makes it easy to diagnose in Actions history. Well done.

---

### Bug 2 — Fixed ✓

Option B was chosen: the download target is now `site/${module}/${BRANCH}`, so artifacts land at `vv_can/development/…` — exactly matching the step summary links. The header comment in each V&V workflow was updated to point to `publish-vv-reports.yml` and the live URL format is correct. Checked `vv-uds.yml`; consistent.

---

### Minor from round 1 — Fixed ✓

Header comment in `vv-uds.yml` (and siblings) now reads:

    #   2. GitHub Pages — published by publish-vv-reports.yml after every
    #       triggered V&V workflow on the SHA completes successfully.

Accurate and concise.

---

### Additional observations (non-blocking)

- The sync warning comment (`# NOTE: the names below must stay in sync with…`) near the `on.workflow_run.workflows` list is a good defensive touch — future contributors renaming a workflow will see the impact immediately.
- `keep_files: true` on the deploy step correctly leaves untouched modules' Pages content intact when only a subset of workflows ran. This is the key invariant that makes the partial-trigger design safe.
- `ran_count=0` exit path is correct; it should never fire in practice (the `if:` condition on the job already requires `conclusion == 'success'` from one of the five workflows), but it is a valid safety net.

Safe to merge into `development`.

erycaaf (approved):
LGTM

